PR #1, opened 2020-06-11
PT-241, PT-242, PT-243, PT-245 | Project skeleton

jorilindell:
1. Initialize new next.js project
2. Add eslint
3. Add jest testing support
4. Add i18n translation support


src/server.js
```diff
@@ -0,0 +1,22 @@
+import express from 'express';
+import next from 'next';
+import nextI18NextMiddleware from 'next-i18next/middleware';
+
+import nextI18next from './i18n.ts';
+
+const port = process.env.PORT || 3000;
```
badrange, line 7: Could be nice to default to another port than 3000 just because we have several projects running on 3000..

badrange: But it might not matter since I think another port will be autoselected

badrange:
PS: Here is a list of things the documentation should contain, among those are CONTRIBUTING.md: 

https://dev.hel.fi/documentation

PR #8, opened 2020-07-01
PT-297 | Event page routing

jorilindell:
1. Generic error page
2. Not found page
3. PageWrapper component to show common meta data
4. Routing to event details page

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -7,7 +7,6 @@ import React from 'react';
 import { Provider } from 'react-redux';
 
 import '../assets/styles/main.scss';
-import { IS_CLIENT } from '../constants';
 import withApollo from '../domain/app/apollo/configureApollo';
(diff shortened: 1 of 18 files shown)


src/pages/_app.tsx
```diff
@@ -7,16 +7,30 @@ import React from 'react';
...
+import { appWithTranslation, i18n } from '../i18n';
 
 interface Props {
   apollo: ApolloClient<NormalizedCacheObject>;
 }
 
 class MyApp extends App<Props> {
+  componentDidMount() {
+    // Change <html>'s language on languageChanged event
+    if (IS_CLIENT) {
```
Sartonon, line 23: componentDidMount is not going to be called server side, so IS_CLIENT is redundant here

jorilindell: So true. Removed redundant IS_CLIENT check

PR #9, opened 2020-07-02
Added RouteComponent interface and jest env variable to setupTest.ts

Sartonon:


Sartonon:
@jorilindell Couple changes/suggestions for codebase

PR #13, opened 2020-07-03
PT-254 | Basic event info

Sartonon:
## Description

Add basic event info to event page

### Closes
**[PT-254](https://helsinkisolutionoffice.atlassian.net/browse/PT-254):** 

## Screenshots
Screenshot 2020-07-03 at 13 54 45


src/domain/event/utils.ts
```diff
@@ -55,3 +56,24 @@ export const getEventStartTimeStr = (
...
+  event: EventFieldsFragment | undefined | null,
+  locale: Language
+) =>
+  event
+    ? {
+        isEventFree: isEventFree(event),
+        eventName: getLocalisedString(event?.name, locale),
+        shortDescription: getLocalisedString(event?.shortDescription, locale),
+        description: getLocalisedString(event?.description, locale),
+        imageUrl: event?.images?.[0]?.url,
```
jorilindell, line 71: Maybe could use placeholder image as default in case that image doesn't exist. You can get placeholder image by using getEventPlaceholderImage util functton

src/domain/event/utils.ts
```diff
@@ -55,3 +56,24 @@ export const getEventStartTimeStr = (
  */
 export const isEventFree = (event: EventFieldsFragment): boolean =>
   Boolean(event.offers.find((item) => item.isFree)?.isFree);
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const getEventFields = (
+  event: EventFieldsFragment | undefined | null,
```
jorilindell, line 62: This helper function is very welcome

src/domain/event/eventBasicInfo/EventBasicInfo.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <div className={styles.infoRight}>
+        {isEventFree && (
+          <div>
+            <IconTicket />
+            <p>{t('event:eventCard.free')}</p>
+          </div>
+        )}
+        <div>
+          <IconInfoCircle />
+          <p>Is there a field for this link? :)</p>
```
jorilindell, line 48: You can find this url from infoUrl {en, fi, sv} field?

src/domain/event/eventBasicInfo/EventBasicInfo.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <section className={styles.eventBasicInfo}>
+      <div className={styles.descriptionPart}>
+        <h1>{eventName}</h1>
+        <p className={styles.shortDescription}>{shortDescription}</p>
+        <p>{description}</p>
+        <EventKeywords event={event} />
+      </div>
+      <div className={styles.infoRight}>
+        {isEventFree && (
+          <div>
```
jorilindell, line 41: Maybe could add a TODO comment that we have to add price information after it's been implemented to palvelutarjotin-admin

src/domain/event/eventBasicInfo/EventBasicInfo.tsx
Sartonon: Thanks. Missed that one :)

src/domain/event/utils.ts
Sartonon: Yeah, we can add more fields as needed, and won't be performance event though it returns many fields that are not always needed

PR #15, opened 2020-07-06
Enrolment form

jorilindell:
### Description
Implements form to create new enrolments

### Closes:
https://helsinkisolutionoffice.atlassian.net/browse/PT-321

### Additional info
Can be tested locally (when using https://palvelutarjotin-api.test.kuva.hel.ninja/graphql as API)  on http://localhost:3000/fi/event/palvelutarjotin:afzii4ixmu/enrolment/create?occurrences=T2NjdXJyZW5jZU5vZGU6Nw==


public/static/locales/en/enrolment.json
```diff
@@ -9,6 +52,20 @@
     "noSelectedOccurrences": "No selected event times"
   },
   "pageTitle": "Enrol for the event",
+  "studyLevel": {
```
Sartonon, line 55: Would usage of https://www.i18next.com/translation-function/plurals#interval-plurals avoid some repetition here?

src/common/components/form/fields/DropdownField.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+    setTimeout(() => {
+      // Automatically call onBlur event to make formik set touched value to true
+      onBlur({
+        target: {
+          id: name,
+        },
+      });
+    });
+  };
```
Sartonon, line 59: Maybe empty line here

public/static/locales/en/enrolment.json
jorilindell: Started to use i18next-intervalplural-postprocessor here. Needed some configuration and new logic when creating grade level options.

PR #18, opened 2020-07-08
PT-322 | Saving new enrolment

jorilindell:
### Description 
Saving new enrolment

### Closes 
PT-322

### Additional info
Can be tested on http://localhost:3000/fi/event/palvelutarjotin:afzii4ixmu/enrolment/create?occurrences=T2NjdXJyZW5jZU5vZGU6Nw==


src/domain/enrolment/CreateEnrolmentPage.tsx
```diff
@@ -49,12 +65,65 @@ const CreateEnrolmentPage: React.FC = () => {
...
+      toast(t('enrolment:errors.createFailed'), {
+        type: toast.TYPE.ERROR,
+      });
+    }
   };
+
+  const error = React.useMemo(() => {
+    if (!isEnrolmentStarted(event))
+      return ENROLMENT_ERRORS.ENROLMENT_NOT_STARTED_ERROR;
+    if (filteredOccurrences.filter((o) => isEnrolmentClosed(o, event)).length)
```
Sartonon, line 118: would this be good place to use .some or .every?

src/domain/enrolment/CreateEnrolmentPage.tsx
```diff
@@ -49,12 +65,65 @@ const CreateEnrolmentPage: React.FC = () => {
...
+      return ENROLMENT_ERRORS.ENROLMENT_CLOSED_ERROR;
+    if (!filteredOccurrences.filter((o) => !hasOccurrenceSpace(o)).length)
```
Sartonon, line 120: would this be good place to use .some or .every?

src/domain/enrolment/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const getNotificationType = (
+  values: EnrolmentFormFields
+): NotificationType | undefined => {
+  if (values.hasEmailNotification && values.hasSmsNotification)
+    return NotificationType.EmailSms;
+  else if (values.hasEmailNotification) return NotificationType.Email;
+  else if (values.hasSmsNotification) return NotificationType.Sms;
+
+  return undefined;
```
Sartonon, line 17: I think returning null here would be better practise. functions return undefined implicitly anyway

PR #28, opened 2020-08-25
PT-346 | Show next occurrences time in event card and filter events without ne…

Sartonon:
## Description

Show next occurrences time in event card and filter events without next occurrence

### Closes
**[PT-346](https://helsinkisolutionoffice.atlassian.net/browse/PT-346):** 

## Additional notes
Something needs to be done with the filtering


jorilindell (approved):
Looks good

PR #29, opened 2020-08-26
Add filter events by language

Sartonon:
## Description

Added filter by language to search form in event page

### Closes
**[PT-391](https://helsinkisolutionoffice.atlassian.net/browse/PT-391):**

### DIFF
--- a/src/domain/events/utils.ts
+++ b/src/domain/events/utils.ts
@@ -37,8 +37,8 @@ export const getInitialValues = (
   return {
     text: getTextFromDict(query, 'text') || '',
     inLanguage: arrayQueryParameterToFormValue(
-      query.inLanguage
-    ) as EVENT_LANGUAGES[],
+      query.inLanguage as EVENT_LANGUAGES
+    ),
   };
 };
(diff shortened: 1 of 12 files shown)


src/domain/events/utils.ts
```diff
@@ -33,9 +36,24 @@ export const getInitialValues = (
 ): EventSearchFormValues => {
   return {
     text: getTextFromDict(query, 'text') || '',
+    inLanguage: arrayQueryParameterToFormValue(
+      query.inLanguage
+    ) as EVENT_LANGUAGES[],
   };
 };
 
+export const arrayQueryParameterToFormValue = <T>(value?: T | T[]): T[] => {
```
Sartonon, line 45: Helper function like this might have been implemented somewhere already?

jorilindell: Maybe you could name this function little bit more generic way e.g. queryParameterToArray and move to src/utils folder? It could be easier to find

jorilindell (approved):
Very good.

PR #31, opened 2020-08-27
Feature/filter events by date

Sartonon:
## Description

Add filter events by date functionality

### Closes
**[PT-390](https://helsinkisolutionoffice.atlassian.net/browse/PT-390):**

### DIFF
--- a/src/domain/events/utils.ts
+++ b/src/domain/events/utils.ts
@@ -1,3 +1,5 @@
+import isValidDate from 'date-fns/isValid';
+
 import { EVENT_LANGUAGES } from '../../constants';
 import {
   EventsQueryVariables,
@@ -28,6 +30,10 @@ export const getEventFilterVariables = (
   include: ['keywords,location'],
   text: getTextFromDict(query, 'text', undefined),
   inLanguage: getTextFromDict(query, 'inLanguage', undefined),
+  start:
+    typeof query.date === 'string' && isValidDate(new Date(query.date))
+      ? new Date(query.date).toISOString()
+      : null,
   ...options,
 });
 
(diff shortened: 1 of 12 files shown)


src/domain/events/utils.ts
```diff
@@ -29,18 +40,34 @@ export const getEventFilterVariables = (
   include: ['keywords,location'],
   text: getTextFromDict(query, 'text', undefined),
   inLanguage: getTextFromDict(query, 'inLanguage', undefined),
+  start: getDateString(query.date),
   ...options,
 });
 
+export const getDateString = (date?: string | string[]): string | null => {
```
jorilindell, line 47: What is the reason to allow string array as date type? I'm just curious

src/domain/events/utils.ts
```diff
@@ -29,18 +40,34 @@ export const getEventFilterVariables = (
...
   query: NodeJS.Dict<string | string[]>
 ): EventSearchFormValues => {
   return {
     text: getTextFromDict(query, 'text') || '',
+    date: getInitialDate(query.date),
     inLanguage: queryParameterToArray(query.inLanguage as EVENT_LANGUAGES),
   };
 };
 
+export const getInitialDate = (date?: string | string[]): Date | null => {
```
jorilindell, line 63: What is the reason to allow string array as date type?

jorilindell (approved):
Looks good to me

src/domain/events/utils.ts
Sartonon: Because the query object can contain string or string array and that is passed as a parameter. This function makes the required checks.

src/domain/events/utils.ts
Sartonon: Same as above. query.date is a query parameter that can be string or string array.

query: NodeJS.Dict<string | string[]>

src/domain/events/utils.ts
Sartonon: I'm open to suggestions if you have some ideas how this query object to form values conversion should be done. Not sure at all if this is good solution

PR #32, opened 2020-09-03
PT-389 | Add place component and filter to events search

Sartonon:
## Description

Add place component and filter to events search

### Closes
**[PT-389](https://helsinkisolutionoffice.atlassian.net/browse/PT-389):** 

## Screenshots
Screenshot 2020-09-03 at 14 02 20

### DIFF
--- a/src/common/components/form/fields/PlaceSelectorField.tsx
+++ b/src/common/components/form/fields/PlaceSelectorField.tsx
@@ -1,0 +1,69 @@
+import { FieldProps } from 'formik';
+import { DropdownProps } from 'hds-react';
+import React from 'react';
+
+import PlaceSelector from '../../../../domain/place/placeSelector/PlaceSelector';
+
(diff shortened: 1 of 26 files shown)


src/common/components/form/fields/PlaceSelectorField.tsx
```diff
@@ -0,0 +1,69 @@
...
+export type Option = {
+  label: string;
+  value: string;
+};
+
+interface Props extends DropdownProps, FieldProps {
+  options: Option[];
+  title: string;
+  inputPlaceholder: string;
+  setFieldValue?: (
```
jorilindell, line 16: Can you import setFieldValue type from formik? If not you could create a re-usable type. I think that this is also defined on DropdownField component

jorilindell (approved):
Very good. Just one really minor improvement idea

PR #36, opened 2020-09-13
PT-371 | Feature/captcha

Sartonon:
## Description

- Add captcha for enrolment form

### Closes
**[PT-371](https://helsinkisolutionoffice.atlassian.net/browse/PT-371):**


jorilindell (approved):
Looks good. You could add capcha key which is working at localhost to .env.local.example file

PR #46, opened 2020-09-30
Improve commit hook

Sartonon:
Improve commit git hook


package.json
```diff
@@ -12,6 +12,7 @@
     "test": "SASS_PATH=./src/assets/styles jest --watch",
     "test:changed": "SASS_PATH=./src/assets/styles jest -o",
     "test:coverage": "SASS_PATH=./src/assets/styles jest --watchAll=false --coverage",
+    "test:staged": "SASS_PATH=./src/assets/styles jest --findRelatedTests",
```
jorilindell, line 15: Do we actually need to set SASS_PATH here? Isn't it already defined in .env?

package.json
```diff
@@ -119,7 +121,14 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm run test:changed"
+      "pre-commit": "lint-staged"
     }
+  },
+  "lint-staged": {
+    "src/**/*.{tsx,ts}": [
+      "npm run test:staged",
```
jorilindell, line 129: Maybe we could use yarn here as well? I don't know does it make any different or is it possible, but it would be nice to be consistent

package.json
```diff
@@ -119,7 +121,14 @@
...
+      "prettier --write",
+      "eslint --fix"
+    ]
```
jorilindell, line 132: Do we also need to run `git add` if eslint or prettier do some changes to files?

jorilindell (approved):
Looks good. Added some minor comments

package.json
Sartonon: Hmm does jest use that?

package.json
Sartonon: yeah maybe yarn is better for consitency

package.json
Sartonon: maybe here has been some update but it complained when I tried to add it. It said that it does the adding for me

package.json
jorilindell: Oh yeah, maybe that was the reason. Maybe there is no need to add support for env if this is only needed variable

Sartonon: yeah but I would be nice to have it in a file. Harder to read these scripts because they are bloated with 1 env variables that is repeated

package.json
jorilindell: Ok, maybe it was some old example which I saw

Sartonon: Yeah I have used git add in the past

PR #48, opened 2020-10-01
Improve tets performance by setting defaultHidden: true

Sartonon:
Improve test performance by configuring defaultHidden to true.

[More info](https://testing-library.com/docs/dom-testing-library/api-queries#byrole)


jorilindell (approved):
Looks good. I though that I approved this allready, but maybe not :D

Sartonon:
> Looks good. I though that I approved this allready, but maybe not :D

It was the other repo :D

PR #51, opened 2020-10-05
Update hds to 0.13.3

Sartonon:
## Description

- Update hds to 0.13.3

- Delete redundant stuff now that hds bundle has commonjs

### Related
**[PT-509](https://helsinkisolutionoffice.atlassian.net/browse/PT-509):**


jorilindell (commented):
Looks good.
Maybe you could remove next-transpile-modules from dependencies if it's not is use anymore?

src/setupTests.ts
```diff
@@ -1,4 +1,2 @@
 import '@testing-library/jest-dom';
 import './tests/initI18n';
-
-process.env.SASS_PATH = './src/assets/styles';
```
jorilindell, line 4: Is there a reason to move SASS_PATH definition back to package.json?

Sartonon: Yeah, that one wasn't actually working, it was some cache that made it look like it was working the other day :(. Spend lot of time to yesterday trying to get that env variable working some other way than in package.json but couldn't.

jorilindell: Ah, ok. In that case, it's better to go with working solution

PR #52, opened 2020-10-06
PT-497 | update hds variables

Sartonon:
## Description

- Update to use updated hds variables

### Closes
**[PT-497](https://helsinkisolutionoffice.atlassian.net/browse/PT-497):**


jorilindell (approved):
Looks good

PR #53, opened 2020-10-09
PT-510 | Accessibility improvements

Sartonon:
## Description

- Event card to links
- Font size and spacing adjustments for table
- Default alt text for event image
- Delete react-i18next dependency and update imports

### Related
**[PT-510](https://helsinkisolutionoffice.atlassian.net/browse/PT-510):**


src/domain/event/eventCard/EventCard.tsx
```diff
@@ -13,10 +15,11 @@ import styles from './eventCard.module.scss';
 
 interface Props {
   event: EventFieldsFragment;
+  link: string;
   onClick?: (id: string) => void;
```
jorilindell, line 19: Could remove unused property

src/domain/event/eventCard/EventCard.tsx
```diff
@@ -26,55 +29,42 @@ const EventCard: React.FC<Props> = ({ event, onClick }) => {
...
-      <div className={styles.contentWrapper}>
-        <div className={styles.titleWrapper}>
-          <div className={styles.title}>{name}</div>
-          <div className={styles.description}>{description}</div>
-        </div>
-        <div className={styles.occurrenceInfoWrapper}>
-          <div className={styles.textWithIcon}>
-            <IconClock />
-            {time}
+    <Link href={link}>
```
jorilindell, line 33: I think that the dynamic routing is not working properly here. When I'm clicking event card i get this error message
Näyttökuva 2020-10-9 kello 14 34 50
And on event page
Näyttökuva 2020-10-9 kello 14 30 34

You can fix this by adding as prop to Link
e.g. 
```
<Link
      href={`${ROUTES.EVENT_DETAILS.replace(':id', '[eventId]')}?eventId=${
        event.id
      }`}
      as={ROUTES.EVENT_DETAILS.replace(':id', event.id)}
    >
```
More info can be found here https://github.com/isaachinman/next-i18next/issues/413

jorilindell (changes requested):

src/domain/event/eventCard/EventCard.tsx
Sartonon: hmm next docs mention that this props is used before version Next.js 9.5.3. Need to check if updating next would help 🤔

Sartonon: And I think I have seen that error evenr before when running locally

jorilindell: The problem existed already before this PR. next-i18next still seems to use 'as' prop internally. Here's source code for that component https://github.com/isaachinman/next-i18next/blob/master/src/components/Link.tsx 

But hopefully just updating next.js to latest version does the magic

Sartonon:
couldn't get this https://github.com/isaachinman/next-i18next/blob/master/examples/simple/next.config.js example working, publicRuntimeConfig was always undefined

Sartonon:
And importing to next.config.js threw error, I will try again :)

jorilindell:
> And importing to next.config.js threw error, I will try again :)

You could try using require. But it really isn't a big deal

Sartonon:
that didn't work but next/config started working, maybe the problem was with the sass config hoc or something

jorilindell (approved):
Very good

PR #54, opened 2020-10-14
PT-411 | Add url slashes to event url if protocol missing

Sartonon:
## Description

- Add //-prefix to url if protocol is missing from it (so that browser knows it is an external url)

### Closes
**[PT-411](https://helsinkisolutionoffice.atlassian.net/browse/PT-411):**

### DIFF
--- a/src/utils/addUrlSlashes.ts
+++ b/src/utils/addUrlSlashes.ts
@@ -2,11 +2,7 @@
 const addUrlSlashes = (url: string): string => {
   const protocolRegex = /^((https?:\/\/)|(\/\/))/i;
   const urlIsMissingProtocol = !protocolRegex.test(url);
-
-  if (urlIsMissingProtocol) {
-    return `//${url}`;
-  }
-  return url;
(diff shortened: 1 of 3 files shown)


src/utils/addUrlSlashes.ts
```diff
@@ -0,0 +1,12 @@
+// add "//" to the beginning of the url if protocol is missing.
+const addUrlSlashes = (url: string): string => {
+  const protocolRegex = /^((https?:\/\/)|(\/\/))/i;
+  const urlIsMissingProtocol = !protocolRegex.test(url);
+
+  if (urlIsMissingProtocol) {
+    return `//${url}`;
+  }
+  return url;
+};
```
jorilindell, line 10: Personally I would prefer ```return urlIsMissingProtocol ? `//${url}` : url``` , but this is really just matter of taste

jorilindell (approved):
Looks good

src/utils/addUrlSlashes.ts
Sartonon: updated to use ternary :)

PR #55, opened 2020-10-21
PT-549 |Update graphql types and code

Sartonon:
## Description

- Update graphql types with codegen

- Update code to work :)

### Closes
**[PT-549](https://helsinkisolutionoffice.atlassian.net/browse/PT-549):**


jorilindell (approved):
Looks good

PR #61, opened 2021-01-17
PT-589 | Focus to top on route changes

Sartonon:
## Description

Focus to top on route changes

### Closes
**[PT-589](https://helsinkisolutionoffice.atlassian.net/browse/PT-589):** 

## Additional notes

Nextjs automarically tries to focus on body element but it needs tabIndex. So workaround is to add it between renders only (might cause issues if it is always there)


src/pages/_app.tsx
```diff
@@ -47,6 +50,18 @@ class MyApp extends App<Props> {
     super.componentDidCatch(error, errorInfo);
   }
 
+  // next focuses body between route changes if it has tabIndex=-1
```
nenonja, line 53: Seem to work when I tested. Just wonder if there would be a chance to write a test for this (at least to check the tabindex is in it's place). But it might be quite hard thing to do on component tests and too corner case for  browser-tests.

nenonja (approved):

src/pages/_app.tsx
nenonja: Hmm one thing is that we should do the same change on palvelutarjotin-admin as well? At least according to the audit report. I dont know if there is separate ticket for it but It could be done now soon.

Sartonon: Yeah noticed that also today.. This ticket was only for Teacher UI. Lets add another ticket if it is missing from the board

Sartonon: We can easily add one test for the focus stuff in browser test later

PR #62, opened 2021-01-22
PT-588: Enrolment mandatory fields marks and notification

melniiv:


src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
```diff
@@ -129,6 +139,8 @@ const EnrolmentForm: React.FC<Props> = ({
               </FormGroup>
               <FormGroup>
                 <Field
+                  required
+                  aria-required="true"
```
Sartonon, line 143: does this need the `="true"` part?

melniiv: i ll test

PR #63, opened 2021-01-23
PT-586 | Improve how seats left is presented

Sartonon:
## Description

- Present seats left so that it is more understandable. Instead of showing 0/100 seats left (which meant 100 places left) show 100/100 seats left

### Closes
**[PT-586](https://helsinkisolutionoffice.atlassian.net/browse/PT-586):** 

## Screenshots

Screenshot 2021-01-23 at 14 40 01

## Additional notes


nenonja (approved):
Code looks good! 
I was trying manually test  a case that I would enrol and reserve one place so that seat count would go -1 but I got some error message (maybe something wrong with local config). Perhaps there could be a e2e browser-test for enrol + seat reservation, but out of scope of this task.

PR #70, opened 2021-02-05
Added price to occurrences accordion.

nikomakela:
PT-733 First price from linked event's offers is shown. Description from the offer and pEvent.paymentInstruction are shown under the price, if they are given.


src/domain/event/occurrences/Occurrences.tsx
```diff
@@ -231,11 +232,16 @@ const Occurrences: React.FC<Props> = ({
...
+    const { placeId, startTime, endTime, linkedEvent, pEvent } = occurrence;
     const date = formatDate(new Date(startTime));
     const time =
       startTime &&
       formatTimeRange(new Date(startTime), new Date(endTime), locale);
+    const offer = linkedEvent?.offers?.[0];
+    const price: string = offer?.price?.[locale] || '';
+    const priceDescription: string = offer?.description?.[locale] || '';
+    const isFree: boolean = offer?.isFree || price === '';
+    const paymentInstruction = pEvent?.paymentInstruction || '';
```
Sartonon, line 244: adding types is adding some unnecessary verbosity, I don't think they are needed 🤔 

if price is string, you could write` !price`  instead of `price === ''`;

also it might be better to use `??` operator instead of `||` even though `||` works fine

src/domain/event/occurrences/Occurrences.tsx
```diff
@@ -251,6 +257,27 @@ const Occurrences: React.FC<Props> = ({
...
+              <IconGlyphEuro />
+            </div>
+            <div>
+              <div className={styles.infoTitle}>
+                <div data-testid="event-price">
+                  {isFree && t('event:occurrenceList.eventIsFree')}
+                  {!isFree && price}
+                </div>
+              </div>
+              {priceDescription !== '' && (
```
Sartonon, line 271: would `priceDescription && (...)` be enough?

nikomakela: I'm not sure, but I thought the left side needs to evaluated as a boolean or otherwise it is rendered. I've had some issues like that earlier.

src/domain/event/occurrences/Occurrences.tsx
nikomakela: All suggestions were good and are now changed.

Sartonon: not pushed yet?

nikomakela:
The API changes should be published before this PR can be used. The paymentInstruction field needs to be removed from API first. Since `event.offers.description` and `event.offers.infoUrl` can be used for same purpose, there is no need for paymentInstruction field and it will be removed as soon as possible.

src/domain/event/occurrences/Occurrences.tsx
Sartonon: I think `priceDescription &&` will work here, see: https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator. If priceDescription is string, the should be no problem. Of course little bit safer way is to force left side be boolean with `!!`

`{!!priceDescription && (`

src/domain/event/occurrences/Occurrences.tsx
```diff
@@ -231,11 +232,27 @@ const Occurrences: React.FC<Props> = ({
...
+    const { placeId, startTime, endTime, linkedEvent } = occurrence;
     const date = formatDate(new Date(startTime));
     const time =
       startTime &&
       formatTimeRange(new Date(startTime), new Date(endTime), locale);
+    const offer = linkedEvent?.offers?.[0];
+    const price = offer?.price?.[locale] || '';
+    const priceDescription = offer?.description?.[locale] || '';
+    const isFree = offer?.isFree ?? !price;
+    const priceInfoUrl = offer?.infoUrl?.[locale] || '';
```
Sartonon, line 244: all those default values after seems to be little bit redundant, I would suggest something like this:

```
    const offer = linkedEvent?.offers?.[0];
    const price = offer?.price?.[locale];
    const priceDescription = offer?.description?.[locale];
    const isFree = offer?.isFree ?? !price;
    const priceInfoUrl = offer?.infoUrl?.[locale];
```

src/domain/event/occurrences/Occurrences.tsx
```diff
@@ -251,6 +268,27 @@ const Occurrences: React.FC<Props> = ({
...
+                </div>
+              </div>
+              {priceDescription !== '' && (
+                <p data-testid="event-priceDescription">{priceDescription}</p>
+              )}
+              {priceInfoUrl !== '' && (
+                <p data-testid="event-priceInfoUrl">
+                  {createLink('', priceInfoUrl)}
+                </p>
+              )}
```
Sartonon, line 289: After deleting some default values above, this can also be

`priceInfoUrl && (...`

src/domain/occurrence/query.ts
```diff
@@ -17,6 +17,26 @@ export const QUERY_OCCURRENCE = gql`
...
+          fi
+          sv
+          en
+        }
+        infoUrl {
+          fi
+          sv
+          en
+        }
+      }
```
Sartonon, line 38: can replaced with offerFields fragment?

```
offers {
        ...offerFields
      }
```

src/domain/event/occurrences/Occurrences.tsx
```diff
@@ -231,11 +232,27 @@ const Occurrences: React.FC<Props> = ({
...
+
+    const createLink = (prefix: string, url: string) => (
+      <>
+        {prefix}
+        {prefix && ' '}
+        <a href={url} rel="noopener noreferrer" target="_blank">
+          {url}
+        </a>
+      </>
+    );
```
Sartonon, line 254: How is this used?

nikomakela: createLink is used in accordion with a price description. It was needed to easily get an clickable url inside a text.

nikomakela:
API changes are already delivered https://github.com/City-of-Helsinki/palvelutarjotin/pull/122

Sartonon (approved):
Great!

src/domain/event/occurrences/Occurrences.tsx
nikomakela: Changed.

src/domain/event/occurrences/Occurrences.tsx
nikomakela: Changed.

src/domain/occurrence/query.ts
nikomakela: I used offerFields, but also improved other parts by using field fragments of image and translations.

PR #72, opened 2021-02-05
PT-685 | Show if event is full in occurrence list

Sartonon:
## Description :sparkles:

- Show if event is full in occurrence list

## Screenshots :camera_flash:

Screenshot 2021-02-05 at 17 49 13


src/domain/enrolment/CreateEnrolmentPage.tsx
```diff
@@ -68,18 +69,27 @@ const CreateEnrolmentPage: React.FC = () => {
...
-          )
-        )
+        ...filteredOccurrences.map((item) => {
+          if (item.seatType === OccurrenceSeatType.ChildrenCount) {
+            return Math.min(
+              item?.maxGroupSize || item.amountOfSeats,
+              item.amountOfSeats - (item.seatsTaken || 0)
+            );
+          }
+          if (item.seatType === OccurrenceSeatType.EnrolmentCount) {
```
nikomakela, line 87: A trivial thing, but this if-statement could be else-if.

Sartonon: I thought that when writing this code but it is 50/50 which style I want to use. To me in this case where we only return something inside the if, else keywords just adds noise.

I prefer this "early return" style.

Sartonon: Tht being said it is not 50/50 for me :D

Sartonon: Did some googling.. some disuccions about that here:

https://dev.to/tyrw/ifelse-or-just-if-1kie

src/domain/occurrence/utils.ts
```diff
@@ -3,13 +3,22 @@ import { isPast, subDays } from 'date-fns';
...
+  if (occurrence.seatType === OccurrenceSeatType.ChildrenCount) {
+    const minGroupSize = occurrence?.minGroupSize || 0;
+    return (
+      minGroupSize < occurrence.amountOfSeats - (occurrence.seatsTaken || 0)
+    );
+  } else if (occurrence.seatType === OccurrenceSeatType.EnrolmentCount) {
+    return occurrence.remainingSeats > 0;
+  }
+
+  return false;
```
Sartonon, line 21: Ofc seems like I haven't been consistent with these if statements then :)

src/domain/enrolment/CreateEnrolmentPage.tsx
```diff
@@ -68,18 +69,27 @@ const CreateEnrolmentPage: React.FC = () => {
...
+            return Math.min(item?.maxGroupSize || 0);
```
nenonja, line 88: Should this be either  `return Math.min(item?.maxGroupSize, 0);` or just `return item?.maxGroupSize || 0`

src/domain/event/EventPage.tsx
```diff
@@ -38,10 +38,17 @@ const EventPage = (): ReactElement => {
...
   >([]);
 
-  const { data: eventData, loading } = useEventQuery({
+  const { data: eventData, loading, refetch: refetchEvent } = useEventQuery({
     variables: { id: eventId as string, include: ['keywords', 'location'] },
   });
 
+  // refetch event data (to get updated occurrences) if enrolment was created
+  React.useEffect(() => {
+    if (enrolmentCreated) {
```
nenonja, line 47: does this get covered in tests?

src/domain/enrolment/CreateEnrolmentPage.tsx
nenonja: One more option is switch + assertUnreachable code check: https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript
Good thing is that if we add new enum type to OccurrenceSeatType then typescript will complain if new type ismissing on switch-case.

src/domain/event/EventPage.tsx
Sartonon: It is not, maybe I will add simpletest to make sure refetch is called on first render...

src/domain/enrolment/CreateEnrolmentPage.tsx
Sartonon: Wow. I have no idea where this Math.min has come from can't really remember writing myself that because this line doesn't make any sense :D

Sartonon: seems like it is some kind of copy paste leftover from the code above, anyway good catch 👍

src/domain/enrolment/CreateEnrolmentPage.tsx
Sartonon: assertUnreachable style is very nice pattern

Sartonon:
```
        ...filteredOccurrences.map((item) => {
          if (item.seatType === OccurrenceSeatType.ChildrenCount) {
            return Math.min(
              item?.maxGroupSize || item.amountOfSeats,
              item.amountOfSeats - (item.seatsTaken || 0)
            );
          }
          if (item.seatType === OccurrenceSeatType.EnrolmentCount) {
            return item?.maxGroupSize || 0;
          }
          return assertUnreachable(item.seatType);
        })
```

That works also

Sartonon: Screenshot 2021-02-10 at 8 49 17
Decided to use switch case because it liiks the cleanest to me :) But left that assertion to default case because otherwise ts complains

src/domain/event/EventPage.tsx
Sartonon: Now fixed

src/domain/event/__tests__/EventPage.test.tsx
```diff
@@ -485,5 +483,51 @@ it('filters occurrence list correctly when sate filters are selected', async ()
   const occurrenceEnrolButtons = screen.getAllByRole('button', {
     name: eventMessages.occurrenceList.enrolOccurrenceButton,
   });
-  expect(occurrenceEnrolButtons).toHaveLength(2);
+  expect(occurrenceEnrolButtons).toHaveLength(1);
+});
+
+describe.only('refetch of event works correctly', () => {
```
nenonja, line 489: only

nenonja: pre-commit hook should get these...

src/domain/occurrence/utils.ts
```diff
@@ -3,13 +3,24 @@ import { isPast, subDays } from 'date-fns';
...
 export const hasOccurrenceSpace = (
   occurrence: OccurrenceFieldsFragment
 ): boolean => {
-  const minGroupSize = occurrence?.minGroupSize || 0;
-  return minGroupSize < occurrence.amountOfSeats - (occurrence.seatsTaken || 0);
+  switch (occurrence.seatType) {
+    case OccurrenceSeatType.ChildrenCount:
+      const minGroupSize = occurrence?.minGroupSize || 0;
+      return (
+        minGroupSize < occurrence.amountOfSeats - (occurrence.seatsTaken || 0)
```
nenonja, line 17: Small thing but it seems that amount of seats left is calculcated in few places, it could be in a util function:
`OccurenceUtils.getAmountOfReservedSeats: (occurence) => occurrence.amountOfSeats - (occurrence.seatsTaken || 0)`

nenonja (commented):
Some comments, I also did not test this.

src/domain/enrolment/CreateEnrolmentPage.tsx
nenonja: Looks nice!

src/domain/occurrence/utils.ts
Sartonon: Hmm would it be `OccurenceUtils.getAmountOfSeatsLeft` instead?

src/domain/event/__tests__/EventPage.test.tsx
Sartonon: fixed and added new eslint plugin

src/domain/occurrence/utils.ts
Sartonon: Fixed

nenonja: That's good name as well

nenonja (approved):
All looks good and approved. But important note that I did not test this, so someone has to do it.

PR #74, opened 2021-02-11
Validate amount of adults after children count is given.

nikomakela:
PT-752 Validate the sum of adults live, before enrolment is submitted.

### DIFF
--- a/src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
+++ b/src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
@@ -397,3 +397,68 @@ test('render and focuses error notification correctly', async () => {
 
   expect(screen.getByRole('textbox', { name: 'Nimi' })).toHaveFocus();
 });
+
+test('amount of adults is limited to 1 when there is 1 seat left after children', async () => {
+  render(<CreateEnrolmentPage />, {
+    mocks: mock3,
+    query: { eventId: eventId, occurrences: occurrenceIds },
+  });
+  await waitFor(() => {
+    expect(screen.getByLabelText(/lapsia/i)).toBeInTheDocument();
+  });
+  userEvent.type(screen.getByLabelText(/lapsia/i), '19');
+  userEvent.type(screen.getByLabelText(/aikuisia/i), '2');
+  userEvent.tab();
+  await waitFor(() => {
+    expect(
+      screen.getByText(
+        /Arvon tulee olla enintään 1 yhdessä lasten lukumäärän kanssa/i
+      )
+    ).toBeInTheDocument();
+  });
+});
+
+test('amount of adults is limited to 0 when children fills the group in whole', async () => {
+  render(<CreateEnrolmentPage />, {
+    mocks: mock3,
+    query: { eventId: eventId, occurrences: occurrenceIds },
+  });
+  await waitFor(() => {
+    expect(screen.getByLabelText(/lapsia/i)).toBeInTheDocument();
+  });
+  userEvent.type(screen.getByLabelText(/lapsia/i), '20');
+  userEvent.type(screen.getByLabelText(/aikuisia/i), '2');
+  userEvent.tab();
+  await waitFor(() => {
+    expect(
+      screen.getByText(
+        /Arvon tulee olla enintään 0 yhdessä lasten lukumäärän kanssa/i
+      )
+    ).toBeInTheDocument();
+  });
+});
+
+test('adults should not validated against max group limit when there are already too many children', async () => {
+  render(<CreateEnrolmentPage />, {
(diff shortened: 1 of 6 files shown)


src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
```diff
@@ -397,3 +397,60 @@ test('render and focuses error notification correctly', async () => {
...
+  });
+  userEvent.type(screen.getByLabelText(/lapsia/i), '20');
+  userEvent.type(screen.getByLabelText(/aikuisia/i), '2');
+  await act(wait);
+  expect(
+    screen.getByText(
+      /Arvon tulee olla enintään 0 yhdessä lasten lukumäärän kanssa/i
+    )
+  ).toBeInTheDocument();
+});
```
nikomakela, line 437: Couldn't get these to work. I don't even know what was wrong. They should mimic these: 
image

src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
```diff
@@ -397,3 +397,60 @@ test('render and focuses error notification correctly', async () => {
...
+  render(<CreateEnrolmentPage />, {
+    mocks: mock3,
+    query: { eventId: eventId, occurrences: occurrenceIds },
+  });
+  await waitFor(() => {
+    expect(screen.getByLabelText(/lapsia/i)).toBeInTheDocument();
+  });
+  userEvent.type(screen.getByLabelText(/lapsia/i), '21');
+  userEvent.type(screen.getByLabelText(/aikuisia/i), '2');
+  await act(wait);
```
nikomakela, line 449: Is there any better alternatives? I tried fireEvents inside an act also, but they didn't work. Example here: https://duncanleung.com/testing-validation-schema-formik-form/

src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
Sartonon: Testing can be tricky at first. I will check tomorrow

Sartonon: Issue with the tests here is that the input where error should appears is not unfocused at any moment, which causes it to go to touched state, and then the error appears.

Can be fixed by just submitting the form or just using userEvent.tab() to focus out of the input.

```
  // userEvent.tab();

  userEvent.click(
    screen.getByRole('button', { name: 'Lähetä ilmoittautuminen' })
  );

  await waitFor(() => {
    expect(
      screen.getByText(
        /Arvon tulee olla enintään 1 yhdessä lasten lukumäärän kanssa/i
      )
    ).toBeInTheDocument();
  });
```

src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
Sartonon: Formik does validation is asynchronous manner so this is perfect place to use the waitFor function.

```
  await waitFor(() => {
    expect(
      screen.getByText(
        /Arvon tulee olla enintään 20/i
      )
    ).toBeInTheDocument();
  });
```

Sartonon: And of course add `userEvent.tab()` to really test this:

```
 expect(
    screen.queryByText(
      /Arvon tulee olla enintään \d yhdessä lasten lukumäärän kanssa/i
    )
  ).toBeNull();
```

Atm I think it passes always because the input is never unfocused

Sartonon: Also `.toBeNull()` -> `.not.toBeInTheDocument()`

Sartonon: and when using `.toBeInTheDocument()` it would be better to use screen.queryBy function because it return null if it doesn't find the element. `getBy` functions throws error itself if it doesn't find the element

Sartonon (commented):
Looks good. My suggestions should get the tests working and then this is approvable :)

src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
nikomakela: worked!

src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
nikomakela: worked! Thank you!

Sartonon (approved):
Looks good now!

PR #77, opened 2021-02-15
Additional information can now be set as mandatory.

nikomakela:
PT-689 PT-767. Enrolment form's additional information field is optional only when a provider allows it to be so.

Provider's UI implementation: https://github.com/City-of-Helsinki/palvelutarjotin-admin/pull/147
Main task: https://helsinkisolutionoffice.atlassian.net/browse/PT-689


nikomakela:
Opened as a draft, because it still needs some test automation.

src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
```diff
@@ -74,6 +83,15 @@ export default Yup.object().shape({
...
           ),
+        extraNeeds: Yup.string().when([], () => {
+          /* 
+          A little Yup-hack with an empty when-block, 
+          since the isMandatoryAdditionalInformationRequired is not part of studyGroup,
+          it needs to be fetched from 1st level when-block so it would be populated.
+          */
+          if (!isMandatoryAdditionalInformationRequired) return Yup.string();
+          return Yup.string().required(VALIDATION_MESSAGE_KEYS.STRING_REQUIRED);
+        }),
```
Sartonon, line 94: Would the following work?

```
  extraNeeds: isMandatoryAdditionalInformationRequired
          ? Yup.string().required(VALIDATION_MESSAGE_KEYS.STRING_REQUIRED)
          : Yup.string(),
```

src/domain/event/utils.ts
```diff
@@ -65,6 +65,15 @@ export const getEventStartTimeStr = (
 export const isEventFree = (event: EventFieldsFragment): boolean =>
   Boolean(event.offers.find((item) => item.isFree)?.isFree);
 
+/**
+ * Check if the additional information is requested as mandatory.
+ * @param eventData
+ * @return {boolean}
+ */
```
Sartonon, line 72: Do these 3 lines add something? these are already documented with typescript. I know these are used somewhere in the codebase but I think these are not needed

src/domain/event/utils.ts
```diff
@@ -65,6 +65,15 @@ export const getEventStartTimeStr = (
...
+export const isMandatoryAdditionalInformationRequired = (
+  event: EventFieldsFragment
+): boolean => Boolean(event.pEvent?.mandatoryAdditionalInformation);
```
Sartonon, line 75: Seems little bit strange to me that this is abstracted to another function, is it really necessary? 🤔

Sartonon (commented):
Looking pretty good :) Couple questions added

src/domain/event/utils.ts
nikomakela: it's not necessary, but it was consistent, since the isFree had similar approach. Also, I feel that pEvent.mandatoryAdditionalInformation is a little bad name for a boolean field (it looks a text field to me 🧐), even now when we are using typed variables. Sure it adds a trivial method in a code base, so in that sense, should we leave it out. 

So which approach will we choose?

src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
nikomakela: I need to try that. Couldn't think so simple :D

nikomakela: it works!

src/domain/event/utils.ts
nikomakela: Because we are coding with TS and have types there, I can agree when the function is as simple as this, or like the other functions in this file. I removed all the manual API-doc comments there - they were just noise.

src/domain/event/utils.ts
nikomakela: I removed it because it was used only here

Sartonon: Ofc there is no right answer but to me I think helper function wasn't needed

Sartonon (approved):
Looking good!

PR #78, opened 2021-02-17
Maximum group size validation made more consistent

nikomakela:
The enrolment button is clickable when there is space for the minimum group size. In other words, the minimum group size can now be less than or equal to the untaken seats (the remaining space) in occurrence.

Added multiple validations which are all described in the Jira task: https://helsinkisolutionoffice.atlassian.net/browse/PT-772?focusedCommentId=19250

This PR removes some work done in https://helsinkisolutionoffice.atlassian.net/browse/PT-752 and https://helsinkisolutionoffice.atlassian.net/browse/PT-685


Sartonon:
Screenshot 2021-02-18 at 19 01 17
Wondering if 0 teachers is possible?

nikomakela:
> Wondering if 0 teachers is possible?

We talked about that with Pirjo (PO) and Tomi and they thought it's a good thing that it can be left to zero, although it is not common.

src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
```diff
@@ -398,66 +398,94 @@ test('render and focuses error notification correctly', async () => {
...
+    userEvent.type(screen.getByLabelText(/lapsia/i), '19');
+    userEvent.type(screen.getByLabelText(/aikuisia/i), '18');
+    userEvent.tab();
+    await waitFor(() => {
+      expect(
+        screen.getByText(
+          /Arvon tulee olla enintään 2 yhdessä aikuisten lukumäärän kanssa/i
+        )
+      ).toBeInTheDocument();
+    });
```
Sartonon, line 460: Some common helper test funciton could be constructed to avoid this repetition. Maybe pass those error messages and amount of kids and adults as parameter

nikomakela: I'm now creating the form in a common function: 

```
    createEnrolmentForm('19', '18');
    await waitFor(() => {
      expect(
        screen.getByText(
          /Arvon tulee olla enintään 2 yhdessä aikuisten lukumäärän kanssa/i
        )
      ).toBeInTheDocument();
    });
```

src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
```diff
@@ -398,66 +398,74 @@ test('render and focuses error notification correctly', async () => {
...
 
-test('amount of adults is limited to 0 when children fills the group in whole', async () => {
-  render(<CreateEnrolmentPage />, {
-    mocks: mock3,
-    query: { eventId: eventId, occurrences: occurrenceIds },
-  });
-  await waitFor(() => {
-    expect(screen.getByLabelText(/lapsia/i)).toBeInTheDocument();
+  test('one field is greater than the max group size and another one is (still) empty', async () => {
+    createEnrolmentForm('21', '');
```
Sartonon, line 430: createEnrolment calls are missing await and I'm getting this error:

Screenshot 2021-02-19 at 14 42 04

src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
```diff
@@ -398,66 +398,74 @@ test('render and focuses error notification correctly', async () => {
...
+  };
+
+  test('both of the fields are greater than the max group size', async () => {
+    createEnrolmentForm('21', '22');
+    await waitFor(() => {
+      expect(
+        screen.getAllByText(
+          /Lasten ja aikuisten yhteislukumäärän tulee olla enintään 20/i
+        )
+      ).toHaveLength(2);
```
Sartonon, line 425: I was thinking if also these error messages could be passed as parameters but this one of course causes some problems.

src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
```diff
@@ -398,66 +398,74 @@ test('render and focuses error notification correctly', async () => {
...
+    });
+    expect(screen.getByText(/Tämä kenttä on pakollinen/i)).toBeInTheDocument();
   });
-  userEvent.type(screen.getByLabelText(/lapsia/i), '20');
-  userEvent.type(screen.getByLabelText(/aikuisia/i), '2');
-  userEvent.tab();
-  await waitFor(() => {
+
+  test('both the fields are valid as a single, but the total is greater than the maximum group size', async () => {
+    createEnrolmentForm('19', '18');
```
Sartonon, line 442: One more minor improvement would be to pass one object instead of 2 parameters. Here it is not very evident what these parameters mean and you have to check the function declaration and remember (or hovering in vscode)

```
createEnrolment({ childrenCount: '19', adultCount: '18' })
```

just a small suggestion, no need to this change if you disagree :)

nikomakela: It would also make a lot of repeat 😉. The createEnrolment is declared inside the describe-function of max group size validation, so I think I will leave it. Because the values are summed, it doesn't even make much difference in which order them are used.

Sartonon: fair enough :)

src/domain/enrolment/__tests__/CreateEnrolmentPage.test.tsx
nikomakela: True that! Strange that I didn't get this error highlighted. However, I think it's now fixed.

Sartonon (approved):
Great!

PR #79, opened 2021-02-24
Added a validation against minimum  group size validation.

nikomakela:
PT-798 Added the missing validation against minimum  group size validation. If the total count in under minimum required, same validation message will appear under both the study group count fields.

Issue: https://helsinkisolutionoffice.atlassian.net/browse/PT-798

Other validation cases here (in a linked task): https://helsinkisolutionoffice.atlassian.net/browse/PT-772?focusedCommentId=19250

### DIFF
--- a/src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
+++ b/src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
@@ -39,26 +39,45 @@ export default Yup.object().shape({
       schema: Yup.ObjectSchema
     ) => {
       const validateSumOfSizePair = (
-        sizePair: number,
+        sizePair: number | undefined,
         schema: Yup.NumberSchema,
-        fieldLimitValidationMessage: string,
-        totalLimitValidationMessage: string
+        totalMinLimitValidationMessage: string,
+        fieldMaxLimitValidationMessage: string,
+        totalMaxLimitValidationMessage: string
       ): Yup.NumberSchema => {
-        // Validate a single field against the max size.
+        // Validate a single field against the total min and max sizes.
         // The used validation error message will be the same for both the fields.
         // This is also preventing negative param.max to occur in validation.
-        if (!sizePair || sizePair < 0 || sizePair > maxGroupSize) {
-          return schema.max(maxGroupSize, (param) => ({
-            max: param.max,
-            key: totalLimitValidationMessage,
-          }));
+        if (
+          !sizePair ||
+          sizePair < 0 ||
+          sizePair < minGroupSize ||
+          sizePair > maxGroupSize
+        ) {
+          return (
+            schema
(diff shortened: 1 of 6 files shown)


nikomakela:
Test shots from PR-server:  

image

image

image

src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
```diff
@@ -41,24 +41,43 @@ export default Yup.object().shape({
...
+          sizePair < 0 ||
+          sizePair < minGroupSize ||
+          sizePair > maxGroupSize
+        ) {
+          return (
+            schema
+              // Min-limit is current a gap to minimum group size.
+              .min(
+                sizePair != null ? minGroupSize - sizePair : minGroupSize,
+                (param) => ({
```
Sartonon, line 62: minor but params not used and can be deleted :)

src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
```diff
@@ -41,24 +41,43 @@ export default Yup.object().shape({
...
+          !sizePair ||
+          sizePair < 0 ||
+          sizePair < minGroupSize ||
+          sizePair > maxGroupSize
+        ) {
+          return (
+            schema
+              // Min-limit is current a gap to minimum group size.
+              .min(
+                sizePair != null ? minGroupSize - sizePair : minGroupSize,
```
Sartonon, line 61: typings suggest that sizePair cant be null or undefined so this logic is little bit strange

Sartonon: Typings could be improved. `sizePair: number | undefined`?

Sartonon (commented):
Must admit that this logic is hard to grasp but code can't always be simple. Fortunately we have good tests and this seems to work :D

nikomakela:
> Must admit that this logic is hard to grasp but code can't always be simple. Fortunately we have good tests and this seems to work :D

I agree! :) All this could be changed to a simpler solution where there would only be same validation messages for both the fields about not met min or max group size. However, I think that it would not be as informative for the user. But that's my suggestion for the next iteration of this validation saga :D

src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
nikomakela: Updated.

src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
nikomakela: Removed.

Sartonon:
Hopefully this validation is enough for now :D

Sartonon (approved):
Very good! 👍

PR #80, opened 2021-02-24
PT-619 | Add categorisation section to event page

Sartonon:
## Description :sparkles:

-  Add categorisation section to event page

### DIFF
--- a/src/domain/event/eventBasicInfo/EventCategorisation.tsx
+++ b/src/domain/event/eventBasicInfo/EventCategorisation.tsx
@@ -1,0 +1,76 @@
+import * as React from 'react';
+
+import {
+  EventFieldsFragment,
+  LocalisedFieldsFragment,
+} from '../../../generated/graphql';
+import useLocale from '../../../hooks/useLocale';
+import { useTranslation } from '../../../i18n';
+import getLocalizedString from '../../../utils/getLocalisedString';
+import { getEventFields, getRealKeywords } from '../utils';
+import styles from './eventBasicInfo.module.scss';
+
+export const categorisationSectionTestId = 'categorisation-section-testid';
+
+const EventCategorisation: React.FC<{
+  event: EventFieldsFragment | null | undefined;
+}> = ({ event }) => {
+  const locale = useLocale();
+  const realKeywords = getRealKeywords({ event });
+  const { t } = useTranslation();
+  const { categories, activities, audience } = getEventFields(event, locale);
+
+  const arrayToText = (
+    items: { name?: LocalisedFieldsFragment | null }[] | undefined
+  ) => {
+    return (
+      items
+        ?.map((item) => getLocalizedString(item.name || {}, locale))
+        .filter(Boolean)
+        .sort()
+        .join(', ') || '-'
+    );
+  };
+
+  return (
+    <div
+      className={styles.eventCategorisation}
+      data-testid={categorisationSectionTestId}
+    >
+      <div className={styles.categorisationRow}>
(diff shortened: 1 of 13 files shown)


src/domain/event/eventBasicInfo/EventCategorisation.tsx
```diff
@@ -0,0 +1,76 @@
...
+    items: { name?: LocalisedFieldsFragment | null }[] | undefined
+  ) => {
+    return (
+      items
+        ?.map((item) => getLocalizedString(item.name || {}, locale))
+        .filter(Boolean)
+        .sort()
+        .join(', ') || '-'
+    );
+  };
```
nikomakela, line 33: Could this be a domain util or something?

src/domain/event/utils.ts
```diff
@@ -125,3 +130,17 @@ export const orderOccurrencesByDate = (
...
+  const { additionalCriteria, categories } = eventData.event || {};
+  return eventData?.event?.keywords.filter((keyword) => {
+    return !(
+      categories?.find((category) => category.id === keyword.id) ||
+      additionalCriteria?.find(
+        (additionalCriteria) => additionalCriteria.id === keyword.id
+      )
+    );
+  });
+};
```
nikomakela, line 146: This could at least include a description. It's hard to understand what the "real keywords" are and what is their purpose - Where are the used or needed?

Wouldn't it be faster and better looking if the filter would be called twice, like:

```
  return eventData?.event?.keywords
    ?.filter((keyword) => !(categories?.find((category) => category.id === keyword.id)))
    ?.filter((keyword) => !(categories?.find((additionalCriteria) => additionalCriteria.id === keyword.id)))
```

Also, doesn't `keywords.filter` need a null check also?

Sartonon: Yes I agree

src/domain/event/eventBasicInfo/EventCategorisation.tsx
Sartonon: Hmm I just used this same way like it was used in admin project. It could be domain util function of course :)

src/domain/event/utils.ts
nikomakela: I think the best practise with filter functions is that it should always do only 1 thing... Like with every function :) I've read several articles about this but now I couldn't find any other but this, which ain't very good: https://advancedweb.hu/best-practices-on-how-to-work-with-collections-in-javascript/

Sartonon: Yeah those should definitely be move to separate filters at least

src/domain/event/eventBasicInfo/EventCategorisation.tsx
Sartonon: Moved to utils

src/domain/event/utils.ts
Sartonon: Improved the function by combining those arrays first
```
// Get keywords that are not already included in additionalCriteria or categories.
export const getRealKeywords = (
  eventData: EventQuery
): Keyword[] | undefined => {
  const { additionalCriteria, categories } = eventData.event || {};

  // Combine other keywords into a single array to make filtering simpler
  const otherKeywords = [...(additionalCriteria ?? []), ...(categories ?? [])];

  return eventData?.event?.keywords.filter(
    (keyword) => !otherKeywords?.find((category) => category.id === keyword.id)
  );
};
```

nikomakela (approved):
Looks good to me

PR #81, opened 2021-02-25
Show zero instead of a negative amount as space left in a study group.

nikomakela:
PT-807 Earlier versions of app filled the event with children only. Now when the adults are added to total seats taken calculation, some negative numbers started to appear. This fix returns 0 instead of negative space left.

Before:
image

### DIFF
--- a/src/domain/occurrence/utils.ts
+++ b/src/domain/occurrence/utils.ts
@@ -23,7 +23,13 @@ export const hasOccurrenceSpace = (
 
 export const getAmountOfSeatsLeft = (
   occurrence: OccurrenceFieldsFragment
-): number => occurrence.amountOfSeats - (occurrence.seatsTaken || 0);
+): number => {
+  const amountOfSeatsLeft =
+    occurrence.amountOfSeats - (occurrence.seatsTaken || 0);
+
+  // Don't return negative space left, but a 0 instead
+  return Math.max(0, amountOfSeatsLeft);
+};
 
 export const isEnrolmentStarted = (
   event?: EventFieldsFragment | null


Sartonon (approved):
👍

src/domain/occurrence/utils.ts
```diff
@@ -23,7 +23,13 @@ export const hasOccurrenceSpace = (
 
 export const getAmountOfSeatsLeft = (
   occurrence: OccurrenceFieldsFragment
-): number => occurrence.amountOfSeats - (occurrence.seatsTaken || 0);
+): number => {
+  const amountOfSeatsLeft =
+    occurrence.amountOfSeats - (occurrence.seatsTaken || 0);
+
+  // Don't return negative space left, but a 0 instead
+  return amountOfSeatsLeft < 0 ? 0 : amountOfSeatsLeft;
```
melniiv, line 31: might be as well Math.max(0, amountOfSeatsLeft) ?

melniiv (approved):

src/domain/occurrence/utils.ts
Sartonon: good point 👍

nikomakela: I agree this looks better. Needed to check what internet says about performance: https://stackoverflow.com/q/1232345/784642. I think I update it.

Sartonon: Maybe if were executing this thousands of times in a second regularly we should consider perf implications but not now when it is basically called couple of times :)

Sartonon: Screenshot 2021-02-25 at 11 04 22
Maybe not even then. Executing that 100 million times take under a second on my computer...

Sartonon: well that is acutally 1 billion times :D?

nikomakela: I was just curious 😂

PR #82, opened 2021-02-25
Improved events query.

nikomakela:
PT-797 Removed pEvent and venue fields (but left an id and nextOccurrenceDatetime) to much improve the events query. Single event query is unchanged.

API changes done in https://github.com/City-of-Helsinki/palvelutarjotin/pull/129.

### DIFF
--- a/src/domain/event/eventKeywords/EventKeywords.tsx
+++ b/src/domain/event/eventKeywords/EventKeywords.tsx
@@ -1,6 +1,9 @@
 import React, { ReactElement } from 'react';
 
-import { EventFieldsFragment } from '../../../generated/graphql';
+import {
+  EventFieldsFragment,
+  EventsFieldsFragment,
+} from '../../../generated/graphql';
 import useLocale from '../../../hooks/useLocale';
 import { useTranslation } from '../../../i18n';
 import getLocalisedString from '../../../utils/getLocalisedString';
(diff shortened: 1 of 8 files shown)


nikomakela:
I first had changes to 1 query file only. It passed testing and linter on my machine, but Github was more demanding :) Needed to add typing here and there.

nikomakela:
Pull request https://github.com/City-of-Helsinki/palvelutarjotin-ui/pull/80 might overlap with this quite badly. This branch should be checked against it.

nikomakela:
In a refinement session it was said that this PR needs to be deployed this week before the next week's demo session.

src/domain/event/eventKeywords/EventKeywords.tsx
```diff
@@ -1,14 +1,17 @@
 import React, { ReactElement } from 'react';
 
-import { EventFieldsFragment } from '../../../generated/graphql';
+import {
+  EventFieldsFragment,
+  EventsFieldsFragment,
```
Sartonon, line 5: Hmm this is the downside passing whole event object as a prop and trying to add types 🤔 Hopefully we don't need more of these field fragments in the future :)

Sartonon (approved):
Looks good.

src/domain/event/eventKeywords/EventKeywords.tsx
nikomakela: True. Otherwise I would have managed with 1 file change.

PR #83, opened 2021-02-26
Changed enrolment form notification language label

nikomakela:
PT-721 PT-828 There are so many language fields in enrolment form that they are confusing.


Sartonon (approved):
Good :)

PR #86, opened 2021-03-03
Occurrences place info map links refactored to use new map link component.

nikomakela:
PT-762. PlaceInfo is refactored to use new PlaceMapLink component, which have 2 render variants: a-element and button-element. UsePlaceMapLinks hook is added to help use of the new map link component.

image

### DIFF
--- a/src/domain/event/occurrences/Occurrences.tsx
+++ b/src/domain/event/occurrences/Occurrences.tsx
@@ -29,7 +29,7 @@ import {
   isEnrolmentClosed,
   isEnrolmentStarted,
 } from '../../occurrence/utils';
-import PlaceInfo from '../../place/placeInfo/PlaceInfo';
+import PlaceInfo, { PlaceInfoLinks } from '../../place/placeInfo/PlaceInfo';
 import PlaceText from '../../place/placeText/PlaceText';
 import CalendarButton from '../calendarButton/CalendarButton';
 import DateFilter from '../dateFilter/DateFilter';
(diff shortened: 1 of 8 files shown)


src/domain/event/occurrences/Occurrences.tsx
```diff
@@ -361,4 +286,129 @@ const Occurrences: React.FC<Props> = ({
...
+      {/* TODO: functionality for these buttons */}
+      <CalendarButton event={event} occurrence={occurrence} />
+      <Button iconLeft={<IconLocation />} variant="supplementary">
+        {t('event:occurrenceList.showAllLocationEvents')}
+      </Button>
+      {/* Move map links down a bit show they would be 
+      closer to location info shown in left column  */}
+      <div style={{ height: '50px' }}></div>
+      {mapLinks.map((link: PlaceMapLinkEntryProps) => (
+        <PlaceMapLink {...link} variant="button" />
```
Sartonon, line 391: Missing key props here. Getting warning: "Warning: Each child in a list should have a unique "key" prop."

nikomakela: This part is still under work, because I need to add the link description there. That key I need anyway 👍

src/domain/event/occurrences/Occurrences.tsx
```diff
@@ -361,4 +286,129 @@ const Occurrences: React.FC<Props> = ({
...
+    <>
+      {/* TODO: functionality for these buttons */}
+      <CalendarButton event={event} occurrence={occurrence} />
+      <Button iconLeft={<IconLocation />} variant="supplementary">
+        {t('event:occurrenceList.showAllLocationEvents')}
+      </Button>
+      {/* Move map links down a bit show they would be 
+      closer to location info shown in left column  */}
+      <div style={{ height: '50px' }}></div>
+      {mapLinks.map((link: PlaceMapLinkEntryProps) => (
```
Sartonon, line 390: Should we use same approach as in PlaceInfo that has PlaceInfoLinks component?

nikomakela: That's coming:
```
  const renderOccurrenceActions = (
    <>
      {/* TODO: functionality for these buttons */}
      <CalendarButton event={event} occurrence={occurrence} />
      <Button iconLeft={<IconLocation />} variant="supplementary">
        {t('event:occurrenceList.showAllLocationEvents')}
      </Button>
      {/* Move map links down a bit show they would be 
      closer to location info shown in left column  */}
      <div style={{ height: '50px' }}></div>
      <PlaceInfoLinks
        id={placeId || eventLocationId}
        language={locale}
        variant="button"
      />
    </>
  );
```

nikomakela:
Current result image

The info will be removed from left column (from the original place). So, when ready, it will be shown only in right column.

nikomakela:
I'm not very happy with this: 
image

It didn't come up with the given design.

Should we override the focus background from all the buttons in right column or try to add the description text inside the button? Any other ideas? Appending it inside the button can be tricky, because HDS styling the button quite heavily.

src/domain/event/occurrences/Occurrences.tsx
```diff
@@ -361,4 +282,128 @@ const Occurrences: React.FC<Props> = ({
...
+  const renderOccurrenceActions = (
+    <>
+      {/* TODO: functionality for these buttons */}
+      <CalendarButton event={event} occurrence={occurrence} />
+      <Button iconLeft={<IconLocation />} variant="supplementary">
+        {t('event:occurrenceList.showAllLocationEvents')}
+      </Button>
+      {/* Move map links down a bit show they would be 
+      closer to location info shown in left column  */}
+      <div style={{ height: '50px' }}></div>
```
Sartonon, line 382: Might make more sense to wrap PlaceInfoLinks with container div and give that margin-top. But can't say there anything wrong with this, not a must change

src/domain/place/placeInfo/PlaceMapLink.tsx
```diff
@@ -0,0 +1,59 @@
...
+        onClick={() => window.open(url)}
+        iconLeft={<IconMap />}
+        variant="supplementary"
+      >
+        {label}
+      </Button>
+      {description && <p className={styles.linkDescription}>{description}</p>}
+    </>
+  );
+};
```
Sartonon, line 57: I was playing little bit with this code to find out if there is a way to make it little bit more concise and came up with this:

```
const PlaceMapLink: React.FC<PlaceMapLinkProps> = ({
  id,
  label,
  description,
  url,
  variant = 'button',
}) => {
  const Component = { a: 'a', button: Button }[variant];
  const isLink = variant === 'a';

  const props = {
    a: {
      href: url,
      rel: 'noopener noreferrer',
      target: '_blank',
      className: styles.linkEntry,
    },
    button: {
      onClick: () => window.open(url),
      iconLeft: <IconMap />,
      variant: 'supplementary' as const,
    },
  }[variant];

  return (
    <>
      <Component id={id} {...props}>
        {isLink && <IconMap className={styles.linkIcon} />}
        {label}
      </Component>
      {description && <p className={styles.linkDescription}>{description}</p>}
    </>
  );
};
```

Sartonon: Seems to work and event typescript is happy. What do you think? At least description and <> is not repeated twice there :D

nikomakela: > I was playing little bit with this code to find out if there is a way to make it little bit more concise and came up with this:

At least that is a very interesting view 😃 I find it harder to read, although the rendering is more compact. We could ask 3rd opinion!

Sartonon: at least <Button /> and <a /> parts could be renderered(returned) in/from separate function?

Sartonon: the section that actually differs from each other

src/domain/event/occurrences/Occurrences.tsx
nikomakela: I was thinking about it too, but I didn't want to make a new (sub) container for half of the buttons. I thought this would be a cleaner solution, since it does not make the buttons different when comparing them to each other.

Sartonon: OK. This approach is ok by me but now the div can be self closed :) `<div style={{ height: '50px' }} />`

src/domain/place/placeInfo/PlaceMapLink.tsx
nikomakela: I was thinking that I could make a component factory there also :D ..or at least split those 2 solutions to 2 different components.

nikomakela: @Sartonon I took your idea and tried to edit it so that there would be less references to variant. Now there is only 1 path selection and 1 render. I'm happy with it like this! 🥳

src/domain/place/placeInfo/PlaceMapLink.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  return (
+    <>
+      <Component id={id} {...props}>
+        {labelPrefix}
+        {label}
+      </Component>
+      {description && <p className={styles.linkDescription}>{description}</p>}
+    </>
+  );
```
Sartonon, line 53: Nice :)

src/domain/place/placeInfo/PlaceMapLink.tsx
Sartonon: Great it is looking good!

Sartonon (approved):
Looking good :)

PR #87, opened 2021-03-09
PT-668 | Add matomo analytics

Sartonon:
## Description :sparkles:

- Add matomo analytics


nikomakela (approved):
Looks good to me

.github/workflows/review.yml
```diff
@@ -30,6 +30,9 @@ jobs:
           DOCKER_BUILD_ARG_NEXT_PUBLIC_SENTRY_DSN: ${{ secrets.GH_SENTRY_DSN }}
           DOCKER_BUILD_ARG_NEXT_PUBLIC_ENVIRONMENT: 'review'
           DOCKER_BUILD_ARG_NEXT_PUBLIC_API_BASE_URL: 'https://palvelutarjotin-api.test.kuva.hel.ninja/graphql'
+          DOCKER_BUILD_ARG_NEXT_PUBLIC_MATOMO_URL_BASE: 'https://analytics.hel.ninja/'
```
nenonja, line 33: does this pollute production analytics with test env analytics? Test environments should have it's own url?

nenonja (commented):
one comment / question

src/domain/app/layout/PageWrapper.tsx
```diff
@@ -12,12 +13,21 @@ interface Props {
...
   const locale = useLocale();
+  const { trackPageView } = useMatomo();
   const { asPath: pathname } = useRouter();
+
   const translatedTitle =
     title !== 'appName'
       ? `${t(title)} - ${t('common:appName')}`
       : t('common:appName');
 
+  // Track page changes when pathnname changes
```
nenonja, line 24: What is the level of analytics we try to achieve? As far as I know, we could track that does users follow certain links, do interactions, how long they stay on the site etc.

Sartonon: I have no idea what kinds of analytics we want.. I think the task was to get this set up. I haven't investogated in matamo what is it doing by default.

Sartonon: This page change was basically copied from events-helsinki-ui

.github/workflows/review.yml
Sartonon: It is not activated in staging for example. Maybe we could delete the env variables from envs where they are not really used

src/domain/app/layout/PageWrapper.tsx
nenonja: Requirements for analytics should come from stakeholders

.github/workflows/review.yml
nenonja: So this can be / should be removed + also from staging.yml ?

nenonja (approved):

PR #88, opened 2021-03-11
PT-617 | Delete empty values from search

Sartonon:
## Description :sparkles:

- Delete empty values from search query string


src/utils/deleteEmptyValuesFromObject.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const deleteEmptyValuesFromObject = (
+  obj: QueryObject
+): Record<string, string | string[]> => {
+  return Object.entries(obj).reduce((acc, [key, value]) => {
+    // If value is null, undefined or empty object/array, don't append
+    if (isNil(value) || isEmpty(value)) return acc;
+    return { ...acc, [key]: value };
+  }, {});
+};
```
nikomakela, line 14: Isn't this already in lodash: `_.pickBy(obj);`? https://stackoverflow.com/a/35871405/784642

src/utils/__tests__/deleteEmptyValuesFromObject.test.ts
```diff
@@ -0,0 +1,32 @@
...
+        test5: ['asd', 'osd'],
+        test6: 'moi',
+      },
+    },
+  ];
+
+  testValues.forEach((t) => {
+    expect(deleteEmptyValuesFromObject(t.param)).toEqual(t.expected);
+  });
+});
```
nikomakela, line 32: Needless if lodash (pickBy) is used instead of custom function.

src/utils/deleteEmptyValuesFromObject.ts
nikomakela: Without lodash that reducer for object entries that you have written would be hot! 👍 ..But then you wouldn't have isNil either :P There are several alternatives introduced here: https://stackoverflow.com/questions/286141/remove-blank-attributes-from-an-object-in-javascript

nikomakela: and note that there are those `_.omit(obj, _.isUndefined)` or `_.omitBy(obj, _.isNil)` alternatives in lodash also! Since you have used isNil, maybe the latter is better?

Sartonon: Nice. I think `_.omitBy(object, _.isNil)` is the way to go as we want to get rid of empty arrays too.

Sartonon: Sorry, we need both _.isNil and isEmpty

Sartonon: `_.omitBy(object, (v) => _.isNil(v) || _.isEmpty(v))`?

src/utils/__tests__/deleteEmptyValuesFromObject.test.ts
Sartonon: We could still abstract that one inside `deleteEmptyValuesFromObject` and leave the tests :)

Sartonon: I think it is easier to understand `deleteEmptyValuesFromObject(object)` than `_.omitBy(object, (v) => _.isNil(v) || _.isEmpty(v))`

Sartonon: hmm or `deleteEmptyPropertiesFromObject(object)`

nikomakela (approved):
good stuff!

PR #89, opened 2021-03-15
PT-870 | Fix event query to include target group

quyenlq:
Screenshot after the fix

Screenshot 2021-03-15 at 18 05 50


Sartonon (approved):
Looking good!

PR #90, opened 2021-03-17
Only upcoming occurrences are listed on event pages

nikomakela:
PT-581 Added an "upcoming" variable to event query's pEvent occurrences.


Sartonon (approved):
Looks good.

PR #91, opened 2021-03-17
Support for new language API.

nikomakela:
PT-897. Pull request https://github.com/City-of-Helsinki/palvelutarjotin/pull/131 for API is needed for this pull request!

### DIFF
--- a/src/domain/occurrence/occurrenceGroupInfo/OccurrenceGroupInfo.tsx
+++ b/src/domain/occurrence/occurrenceGroupInfo/OccurrenceGroupInfo.tsx
@@ -10,7 +10,9 @@ interface Props {
 const OccurrenceGroupInfo: React.FC<Props> = ({ occurrence }) => {
   const { t } = useTranslation();
 
-  const languages = occurrence.languages.map((language) => language.id);
+  const languages = occurrence.languages.edges.map(
+    (edge) => edge?.node?.id ?? ''
+  );
   const amountOfSeats = occurrence?.amountOfSeats;
   const maxGroupSize = occurrence?.maxGroupSize;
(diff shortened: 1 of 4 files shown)


src/domain/occurrence/occurrenceGroupInfo/OccurrenceGroupInfo.tsx
```diff
@@ -10,7 +10,8 @@ interface Props {
 const OccurrenceGroupInfo: React.FC<Props> = ({ occurrence }) => {
   const { t } = useTranslation();
 
-  const languages = occurrence.languages.map((language) => language.id);
+  const languages =
```
Sartonon, line 13: According to type `|| []` seems to be redundant:

```
  const languages = occurrence.languages.edges.map(
    (edge) => edge?.node?.id ?? ''
  );
```

Sartonon (approved):
1 minor comment but approved

src/domain/occurrence/occurrenceGroupInfo/OccurrenceGroupInfo.tsx
nikomakela: Fixed as suggested.

Sartonon (approved):

nikomakela:
Before that tiny fix this PR somehow passed the acceptance test. Now it fails, but it really should fail because the API is not yet deployed. Let's run the acceptance test again when the API is shipped.

nikomakela:
> Before that tiny fix this PR somehow passed the acceptance test. Now it fails, but it really should fail because the API is not yet deployed. Let's run the acceptance test again when the API is shipped.

It get fixed by rerun already 🤔

PR #92, opened 2021-03-18
Rendered event languages on event page

nikomakela:
PT-830. In language field is now included in event query and the languages are listed in category section because it was designed there and the field works exactly the same as keywords. Changes to how the group info with language is shown on occurrence. Added labels for seats taken and language. Some changes to translations.

image

image

### DIFF
--- a/src/domain/occurrence/occurrenceGroupInfo/OccurrenceGroupLanguageInfo.tsx
+++ b/src/domain/occurrence/occurrenceGroupInfo/OccurrenceGroupLanguageInfo.tsx
@@ -1,0 +1,26 @@
+import React from 'react';
+
+import { OccurrenceFieldsFragment } from '../../../generated/graphql';
+import { useTranslation } from '../../../i18n';
+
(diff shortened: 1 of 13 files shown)


src/domain/occurrence/occurrenceGroupInfo/OccurrenceGroupLanguageInfo.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+interface Props {
+  occurrence: OccurrenceFieldsFragment;
+}
+
+const OccurrenceGroupLanguageInfo: React.FC<Props> = ({ occurrence }) => {
+  const { t } = useTranslation();
+
+  const languages =
+    occurrence.languages.edges.map((edge) => edge?.node?.id || '') || [];
```
Sartonon, line 14: Minor but `??` could be used instead of `||`

Sartonon (approved):
Canät test this right now but looks good to me :)

PR #93, opened 2021-03-19
Weekday shown in full in occurrence table.

nikomakela:
PT-851 - "As a teacher I want see weekday at enrolment occurence table."

image

image

image


Sartonon (approved):
Looks good :)

PR #96, opened 2021-03-25
Splitted Occurrences component to multiple new private components.

nikomakela:
PT-911 Occurrences component was a massive function that had issues with hooks. Because React hooks doesn't allow early returns, a bug in occurrences rendering could not be resolved without a great refactoring.

Screenshots of the wanted result and the error log in issue: https://helsinkisolutionoffice.atlassian.net/browse/PT-911

image


Sartonon (approved):
Looking good :)

PR #97, opened 2021-03-25
Email is mandatory but phone number is optional in the enrolment form

nikomakela:
PT-891. Enrolments does not require a phone number anymore.

Added missing required-attributes to emails and names. Removed required-attributes and validations from phone fields. 

Enrolment's email notifications are now mandatory. The checkbox is auto selected and disabled in UI, so it appears as checked but passive. GraphQL mutation does not allow sms only notifications anymore. It should be noted that API still allows them!

### DIFF
--- a/src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
+++ b/src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
@@ -108,8 +108,6 @@ const EnrolmentForm: React.FC<Props> = ({
               </FormGroup>
               <FormGroup>
                 <Field
-                  required
-                  aria-required
                   labelText={t(
                     nameToLabelPath['studyGroup.person.phoneNumber']
                   )}
@@ -207,6 +205,8 @@ const EnrolmentForm: React.FC<Props> = ({
                 <>
                   <FormGroup>
                     <Field
+                      required
+                      aria-required
                       labelText={t(nameToLabelPath['person.name'])}
                       component={TextInputField}
                       name="person.name"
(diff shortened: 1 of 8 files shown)


src/domain/enrolment/utils.ts
```diff
@@ -5,15 +5,10 @@ import {
 } from '../../generated/graphql';
 import { EnrolmentFormFields } from './enrolmentForm/constants';
 
-const getNotificationType = (
-  values: EnrolmentFormFields
-): NotificationType | undefined => {
+const getNotificationType = (values: EnrolmentFormFields): NotificationType => {
```
Sartonon, line 8:
```
ternary might read nicer in this situation :)

return values.hasEmailNotification && values.hasSmsNotification
    ? NotificationType.EmailSms
    : NotificationType.Email;
```

src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
```diff
@@ -232,10 +234,14 @@ const EnrolmentForm: React.FC<Props> = ({
               <div className={styles.divider} />
 
               <h2>{t('enrolment:enrolmentForm.titleNotifications')}</h2>
+              {errors.hasEmailNotification && touched.hasEmailNotification && (
+                <ErrorMessage>{t(errors.hasEmailNotification)}</ErrorMessage>
+              )}
```
Sartonon, line 239: email notifications are always checked and disabled, why is this needed?

src/domain/enrolment/utils.ts
nikomakela: true that. Because of some reason I wanted to keep it easily recoverable / extendable :)

src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
nikomakela: Maybe it is not. I didn't think about it much. I thought there could be a way to achieve it some how...

I will take it off.

src/domain/enrolment/utils.ts
Sartonon: Well it wasn't wrong :)

Sartonon (approved):

PR #98, opened 2021-03-26
PT-365 | Add feedback link

Sartonon:
## Description :sparkles:

- Add feedback link to Footer

### DIFF
--- a/src/domain/app/footer/Footer.tsx
+++ b/src/domain/app/footer/Footer.tsx
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import { Koros } from 'hds-react';
 import React from 'react';
 
-import { PRIVACY_POLICY_LINKS } from '../../../constants';
+import { FEEDBACK_LINKS, PRIVACY_POLICY_LINKS } from '../../../constants';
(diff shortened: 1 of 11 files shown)


src/domain/app/footer/Footer.tsx
```diff
@@ -41,6 +41,10 @@ const Footer = ({ isMobileMenuOpen = false }: Props): React.ReactElement => {
               <Link href="/accessibility">
                 {t('common:accessibilityStatement.title')}
               </Link>{' '}
+              •{' '}
+              <a href={FEEDBACK_LINKS[locale]} target="_blank" rel="noreferrer">
```
melniiv, line 45: should we create a custom components for external link with preconfigured attributes?

Sartonon: That is an option but I can't see many more of these links coming so is there a big benefit? So for time saving (tight budget) should we go with this?

Sartonon: I thought this more and maybe it would be good to have Compoment for external links... It could also always put helper text for screen readers.

melniiv (approved):

PR #99, opened 2021-03-26
Implemented Show all events of this location -button feature.

nikomakela:
PT-835 Show all events of this location -button opens a new window with a place search.

Button simple opens a new tab.

image

### DIFF
--- a/src/domain/event/occurrences/Occurrences.tsx
+++ b/src/domain/event/occurrences/Occurrences.tsx
@@ -374,11 +374,20 @@ const OccurrenceInfo: React.FC<OccurrenceInfoProps> = ({
     </>
   );
 
+  const showAllLocationEvents = () => {
+    const url = `/?places=${placeId || eventLocationId}`;
+    window.open(url);
+  };
+
   const renderOccurrenceActions = (
     <>
       {/* TODO: functionality for these buttons */}
       <CalendarButton event={event} occurrence={occurrence} />
-      <Button iconLeft={<IconLocation />} variant="supplementary" disabled>
+      <Button
+        iconLeft={<IconLocation />}
+        variant="supplementary"
+        onClick={showAllLocationEvents}
+      >
         {t('event:occurrenceList.showAllLocationEvents')}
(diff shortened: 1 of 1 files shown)


src/domain/event/occurrences/Occurrences.tsx
```diff
@@ -374,11 +374,20 @@ const OccurrenceInfo: React.FC<OccurrenceInfoProps> = ({
     </>
   );
 
+  const showAllLocationEvents = () => {
+    const url = `/?places=${placeId || eventLocationId}`;
+    window.open(url);
+  };
```
Sartonon, line 380: Do we want to open this in a new window? If so, button might need text for screen reader that says link opens in a new window.

nikomakela: I don't know, do we? I chose this approach so the user who is enrolling does not lost the enrolment page he was in. Should I change it to `location.href=url`? Would that be better?

Sartonon: This will be caught in next accessibility audit if we open in a new tab without warning the user (screen reader). 

We should as Pirjo / Joakim about this :)

nikomakela: Asked from Joakim. If there can be a button inside the Link-component without wrecking styles I can use it. This doc page says there can https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-function-component.

Sartonon: Great, seems like HDS forwards the ref which is required with that:

https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/packages/react/src/components/button/Button.tsx#L80

Sartonon (approved):
Great :)

src/domain/event/occurrences/Occurrences.tsx
nikomakela: yes, and the passHref was an important parameter.

PR #102, opened 2021-04-08
Don't allow sms notifications if no phone number is given

nikomakela:
PT-880. If no phone number field is filled, sms-notifications checkbox should be disabled and unchecked. If any of the phone number fields is filled, it should be enabled and free to use.

image

image

### DIFF
--- a/src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
+++ b/src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
@@ -52,20 +52,21 @@ const EnrolmentForm: React.FC<Props> = ({
       onSubmit={onSubmit}
       validationSchema={ValidationSchema}
     >
-      {({
-        errors,
-        handleSubmit,
-        touched,
-        submitCount,
-        values: {
+      {({ errors, handleSubmit, touched, submitCount, values }) => {
+        const {
           isSameResponsiblePerson,
           isMandatoryAdditionalInformationRequired,
-        },
-      }) => {
+          studyGroup: {
+            person: { phoneNumber: studyGroupPhoneNumber },
+          },
+        } = values;
         const showErrorNotification = !isEmpty(errors) && !!submitCount;
         const errorLabelKeys = keyify(errors)
           .map((path) => nameToLabelPath[path])
           .filter((i) => i);
+        const hasPhoneNumber = () => {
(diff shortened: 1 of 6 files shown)


src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
```diff
@@ -57,15 +57,43 @@ const EnrolmentForm: React.FC<Props> = ({
...
+        };
+
+        const handleStudyGroupPhoneNumberChange = (
+          e: React.ChangeEvent<HTMLInputElement>
+        ) => {
+          handleChange(e);
+          if (!e.target.value && !phoneNumber) {
+            setFieldValue('hasSmsNotification', false);
+          }
+        };
```
Sartonon, line 95: This way of changing form value based on other form field might cause some problems. Maybe it happens to work without problems in this situation. Heres how creator of Formik recommends to do this:

https://github.com/formium/formik/issues/2204#issuecomment-574207100

Sartonon (commented):
Not sure if we should change the way we change field values based on other field.

Also I'm not sure if current implementation is very user friendly. Can be confusing why the checkbox is disabled.

src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
nikomakela: Some reason why the HDS checkbox checked value did not get cleared or at least the cleared checkbox was not rerendered after the value change. This was the only solution I get it to work. I tried to extend the HDS checkbox field component with a useEffect like that, but I think I used wrong hook for value reading / set.

nikomakela:
> Also I'm not sure if current implementation is very user friendly. Can be confusing why the checkbox is disabled.

I added there a title (tooltip) to tell the reason. I agree it should be more visible, but the HDS checkbox does not have helper texts, so I don't know how the information should be added there. Any suggestions? Should I create a helper text kind of span-element there?

The email notification is currently checked and disabled - the purpose is to indicate that it is automatically checked and it cannot be unchecked. The sms notification field should work the same when there are no any phone numbers given, but since  it's automatically unselected, it should clear any inputs given by an user and the checkbox should be rendered as disabled but unchecked.

Sartonon (approved):
Great! Looks good to me!

PR #103, opened 2021-04-09
Show all events from provider link added to event basic information.

nikomakela:
PT-298 Events can be searched with organisationId. Added organisationId in events query arguments.

Show all events from provider link added to event basic information. Refactored organization information to a new OrganizationInfo -component. Added there a router link that would open a search page with organization parameter. Search by organization is still missing.

The query can now be made with organization -parameter, which should be an organizationId. E.g http://localhost:3000/fi?organization=T3JnYW5pc2F0aW9uTm9kZToy

image

### DIFF
--- a/src/domain/event/__tests__/EventPage.test.tsx
+++ b/src/domain/event/__tests__/EventPage.test.tsx
@@ -636,4 +636,56 @@ describe('refetch of event works correctly', () => {
       ).toBeInTheDocument();
     });
   });
+
+  it('renders a link which leads to all events of the organisation', async () => {
+    renderComponent();
+    await waitForRequestsToComplete();
+    expect(
+      screen.queryByText('Näytä järjestäjän tapahtumat')
+    ).toBeInTheDocument();
+  });
+
+  it('does not render organisation section when organisation is not given', async () => {
+    const eventWithoutOrganisation = {
+      ...eventData,
+      data: {
+        ...eventData.data,
+        event: {
+          ...eventData.data.event,
+          pEvent: {
+            ...eventData.data.event.pEvent,
+            organisation: {
+              ...eventData.data.event.pEvent.organisation,
+              name: '',
+            },
+          },
+        },
+      },
+    };
+    const mocks = apolloMocks.map((m, index) => {
+      const mock = { ...m };
+      if (index === 0) {
+        mock.result = eventWithoutOrganisation as any;
+      }
+      return mock;
+    });
+
+    render(<EventPage />, {
+      mocks: mocks,
+      query: { eventId: eventData.data.event.id },
+      path: `/fi${ROUTES.EVENT_DETAILS.replace(':id', data.id)}`,
+    });
+    await waitForRequestsToComplete();
+    expect(
+      screen.queryByText('Näytä järjestäjän tapahtumat')
+    ).not.toBeInTheDocument();
+
+    [
+      data.contactPersonName,
+      data.contactPersonEmail,
+      data.contactPersonPhoneNumber,
+    ].forEach((text) =>
+      expect(screen.queryByText(text)).not.toBeInTheDocument()
+    );
+  });
 });
--- a/src/domain/event/__tests__/EventPage.test.tsx
+++ b/src/domain/event/__tests__/EventPage.test.tsx
@@ -640,9 +640,12 @@ describe('refetch of event works correctly', () => {
   it('renders a link which leads to all events of the organisation', async () => {
     renderComponent();
     await waitForRequestsToComplete();
-    expect(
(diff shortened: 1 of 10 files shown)


src/domain/event/eventBasicInfo/EventBasicInfo.tsx
```diff
@@ -58,21 +59,54 @@ const EventBasicInfo: React.FC<EventBasicInfoProps> = ({ event }) => {
         )}
         <div>
           <IconFaceSmile />
```
Sartonon, line 61: Should these be inside `{organization && organizationId && (`

Sartonon (commented):
Looking good, couple comments added.

nenonja (commented):
No EventBasicInfo test for new functionality? With / without the link.
Perhaps also query parameter test for the front page too, if it's not too hard to implement.

nikomakela:
> No EventBasicInfo test for new functionality? With / without the link.
> Perhaps also query parameter test for the front page too, if it's not too hard to implement.

Older tests was not found for either of the features. I added some for organisation section. The search query is still untested, but it's a huge process that we need to start some other time.

nikomakela:
Manual testing instruction and result: https://helsinkisolutionoffice.atlassian.net/browse/PT-298?focusedCommentId=38416

src/domain/event/__tests__/EventPage.test.tsx
```diff
@@ -636,4 +636,56 @@ describe('refetch of event works correctly', () => {
       ).toBeInTheDocument();
     });
   });
+
+  it('renders a link which leads to all events of the organisation', async () => {
+    renderComponent();
+    await waitForRequestsToComplete();
+    expect(
```
nenonja, line 643: queryBy should only be used when something is not in the document:  https://testing-library.com/docs/queries/about/
>queryBy...: Returns the first matching node for a query, and return null if no elements match. This is useful for asserting an element that is not present. Throws an error if more than one match is found (use queryAllBy instead if this is OK).
>findBy...: Returns a Promise which resolves when an element is found which matches the given query. The promise is rejected if no element is found or if more than one element is found after a default timeout of 1000ms. If you need to find more than one element, use findAllBy.

This could be more preferred check:
`await screen.findByRole('link', {  name: /näytä järjestäjän tapahtumat/i });`

src/domain/event/__tests__/EventPage.test.tsx
```diff
@@ -636,4 +636,56 @@ describe('refetch of event works correctly', () => {
...
+      return mock;
+    });
+
+    render(<EventPage />, {
+      mocks: mocks,
+      query: { eventId: eventData.data.event.id },
+      path: `/fi${ROUTES.EVENT_DETAILS.replace(':id', data.id)}`,
+    });
+    await waitForRequestsToComplete();
+    expect(
```
nenonja, line 679: `queryByRole('link', { name: ...`

src/domain/event/__tests__/EventPage.test.tsx
```diff
@@ -636,4 +636,56 @@ describe('refetch of event works correctly', () => {
...
     });
   });
+
+  it('renders a link which leads to all events of the organisation', async () => {
+    renderComponent();
+    await waitForRequestsToComplete();
+    expect(
+      screen.queryByText('Näytä järjestäjän tapahtumat')
+    ).toBeInTheDocument();
+  });
```
nenonja, line 646: Could we check also
```
 data.contactPersonName,
      data.contactPersonEmail,
      data.contactPersonPhoneNumber,
```
are present?

nenonja (commented):
Almost there! Some small fixes to the test

src/domain/event/__tests__/EventPage.test.tsx
nenonja: nitpick but use `queryBy` only with `not.toBeInTheDocument` and findBy otherwise. It could be:

```
[
      data.contactPersonName,
      data.contactPersonEmail,
      data.contactPersonPhoneNumber,
    ].forEach(async (text) => await screen.findByText(text);
```
or could this work?
```
await promise.all([
      data.contactPersonName,
      data.contactPersonEmail,
      data.contactPersonPhoneNumber,
    ].map(screen.findByText);
```

nenonja (approved):
Still one comment but can be approved if not bothered to fix it :)

nikomakela:
> Still one comment but can be approved if not bothered to fix it :)

Neither worked, so I continue without changes. With promise it didn't understand the map function and with async-await-findbytext, I get the following:

`Warning: You seem to have overlapping act() calls, this is not supported. Be sure to await previous act() calls before making a new one.`

PR #107, opened 2021-04-19
PT-953: Add remote event categories to the top of dropdown options

melniiv:


nikomakela (approved):
Looks good to me. A test could be added, but maybe it's not so important here now.

PR #112, opened 2021-04-27
PT-964: Restyling, banner hero and footer

melniiv:


nikomakela (approved):
I did a light review and it all looks good to me. We already know that there is a high probability that more changes are requested in near future, so I wouldn't use much time to polish the code - if it is even needed. We have already talked about new front page, changes to search form, etc.

nikomakela:
The hero title text is a very long one. It should be something prettier, but PO and designers can design that.

PR #116, opened 2021-05-03
PT-964: Restyling improvements

melniiv:


src/domain/app/header/Header.tsx
```diff
@@ -19,6 +19,8 @@ const Header: React.FC = () => {
   const toggleMenu = () => setMenuOpen(!menuOpen);
   const closeMenu = () => setMenuOpen(false);
 
+  const navigatinLinksEnabled = false;
```
Sartonon, line 22: small typo: navigatin -> navigation

melniiv: true

Sartonon (approved):

PR #117, opened 2021-05-06
PT-986: Adds filter badges to the search form

melniiv:
## Description :sparkles:
Only provider filter is implemented


nikomakela (approved):
The code looks good to me. Badge could be some how brighter - I mean it might be hard to notice when using the site at the first time. For me it even appears a second or 2 after page load, so it could be more visible.

image

PR #122, opened 2021-05-10
Fix publisher badge

melniiv:

### DIFF
--- a/src/domain/events/eventSearchForm/EventSearchForm.tsx
+++ b/src/domain/events/eventSearchForm/EventSearchForm.tsx
@@ -22,6 +22,8 @@ export type EventSearchFormValues = {
   endDate: Date | null;
   places: string[];
   organisation?: string;
+  organization?: string;
+  organisationId?: string;
 };
 
 const defaultInitialValues: EventSearchFormValues = {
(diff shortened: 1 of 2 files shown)


nikomakela (approved):
The new keys are confusing, but if the API gives them like that, there is not much to do. Maybe it could have some kind of adapter that there would be no keys that stores the same value, e.g organisation & organization, when they both are initialized with the organisation name.

src/domain/events/eventSearchForm/EventSearchForm.tsx
```diff
@@ -22,6 +22,8 @@ export type EventSearchFormValues = {
   endDate: Date | null;
   places: string[];
   organisation?: string;
+  organization?: string;
+  organisationId?: string;
```
Sartonon, line 26: Are you sure this is correct? There is now organization and organisation with z and s?

melniiv: yes, we have a mess with those s and z all over the project. in query string we use z sometimes, schemas have s. and mappers have different values. we basically need to refactor then all places (and backend as well) to make it conventional

melniiv: @Sartonon you left the call when i explained why its a "tricky" solution. so the code is not nice i agree.. but might be temporary working solution.

Sartonon:
I think we should fix those mismatching thing and use the one api uses? Or does API have 2 version of organization too? I think might not be that big of a task because we are useing codegen and typescript

PR #140, opened 2021-08-23
PT-1135 | Fix use event.offer instead of occurrence.linkedEvent to improve api perf

Sartonon:
## Description :sparkles:

- Fix use event.offer instead of occurrence.linkedEvent to improve api perf

### Closes :no_good_woman:

**[PT-1135](https://helsinkisolutionoffice.atlassian.net/browse/PT-1135):**


nikomakela (approved):
Nice!

PR #143, opened 2021-08-31
Search Headless CMS pages

nikomakela:
Pages can be searched from all the pages or just from 1 page's sub pages.

PT-1126.


nikomakela:
First draft version of the search feature:

image

src/headless-cms/components/CmsPageSearch/CmsPageSearch.tsx
```diff
@@ -0,0 +1,179 @@
...
+  PageIdType,
+  usePagesSearchQuery,
+  useSubPagesSearchQuery,
+} from '../../../generated/graphql-cms';
+import { useCMSClient } from '../../cmsApolloContext';
+import styles from './cmspagesearch.module.scss';
+
+const CmsPageSearch: React.FC<{
+  page?: Page | undefined | null;
+}> = ({ page }): JSX.Element => {
```
Sartonon, line 19: Return type no necessarily needed when React.FC is already provided (it already includes it)

src/headless-cms/components/CmsPageSearch/CmsPageSearch.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  // Update the search results
+  React.useEffect(() => {
+    if (!loading && pageData?.page) {
+      const subPages = pageData?.page?.children?.edges
+        ?.map((edge) => edge?.node as Page)
+        .filter((page) => !!page);
+      setSearchResult(subPages ?? []);
+    }
+  }, [searchTerm, pageData, loading]);
```
Sartonon, line 89: is useEffect and separate state for searchResults actually needed here? COuld this be just:

```
  const subPages = pageData?.page?.children?.edges
        ?.map((edge) => edge?.node as Page)
        .filter((page) => !!page);
```

src/headless-cms/components/CmsPageSearch/CmsPageSearch.tsx
```diff
@@ -0,0 +1,179 @@
...
+        onSubmit={(submittedValue) => setSearchTerm(submittedValue)}
+      />
+    </div>
+  );
+
+  if (page) {
+    return searchForm(`Search from ${page.title} page's subpages...`);
+  }
+
+  return searchForm('Search from all the pages...');
```
Sartonon, line 122: Abstracting to a function here seems little bit overkill here 🤔 
```
 <div className={styles.searchForm}>
      <SearchInput
        label="Search"
        helperText={page ? `Search from ${page.title} page's subpages...` : 'Search from all the pages...'}
        searchButtonAriaLabel="Search"
        clearButtonAriaLabel="Clear search field"
        onSubmit={(submittedValue) => setSearchTerm(submittedValue)}
      />
    </div>
```

src/headless-cms/query.ts
```diff
@@ -72,12 +72,23 @@ export const PAGE_QUERY = gql`
...
+    featuredImage {
+      node {
+        mediaItemUrl
+        link
+        altText
+        mimeType
+        title
+        uri
+      }
+    }
```
Sartonon, line 91: Do we need this at all times?

src/headless-cms/components/CmsPageSearch/CmsPageSearch.tsx
nikomakela: Maybe not. I removed it and it seems to work ok.

src/headless-cms/components/CmsPageSearch/CmsPageSearch.tsx
nikomakela: changed to 
```

  const helperText = page
    ? `Search from ${page.title} page's subpages...`
    : 'Search from all the pages...';

  return (
    <div className={styles.searchForm}>
      <SearchInput
        label="Search"
        helperText={helperText}
        searchButtonAriaLabel="Search"
        clearButtonAriaLabel="Clear search field"
        onSubmit={(submittedValue) => setSearchTerm(submittedValue)}
      />
    </div>
  );
```

src/headless-cms/query.ts
nikomakela: I thought it would just be simpler if it is. I also thought that if the page has a featured image, the page-view if anything, should show the picture. Then it would already be included in listing and the details page. Sure it won't be used in menus, but multiple different kind of field fragments are hard to maintain too.

PR #146, opened 2021-09-05
refactor(events-list): New sorters and filters for events.

nikomakela:
Started using only 2 different events filters and sorters. Both are descending by start times, but another uses events from LinkedEvents and another uses occurrences from API.

PT-1150 PT-1154.

### DIFF
--- a/public/locales/en/events.json
+++ b/public/locales/en/events.json
@@ -11,7 +11,7 @@
   },
   "sortOptions": {
     "lastModifiedTimeDesc": "Last modified",
-    "occurrenceStartTimeDesc": "Occurrence date"
+    "startTime": "Topical"
   },
   "search": {
     "buttonClear": "Clear search criteria",
--- a/public/locales/en/events.json
+++ b/public/locales/en/events.json
@@ -10,14 +10,8 @@
     "title": "Events"
   },
   "sortOptions": {
-    "duration": "Duration, ascending",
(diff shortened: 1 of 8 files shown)


nikomakela:
Since the sorting by occurrence data could not be done, the default sort is now event start time and it's text is changed to "Ajankohtaista"

image

Sartonon (approved):

nikomakela:
I don't understand the reason why the language selector cannot be clicked:

This code always throws from `.click(header.languageSelector)` and there is nothing wrong in screen shot - the button is just not clicked!
```
  await t
    .expect(header.languageSelector.exists)
    .ok()
    .click(header.languageSelector)
    .click(header.languageSelectorItemSv)
    .expect(getPathname())
    .eql('/sv');
```

nikomakela:
> I don't understand the reason why the language selector cannot be clicked:
> 
> This code always throws from `.click(header.languageSelector)` and there is nothing wrong in screen shot - the button is just not clicked!
> 
> ```
>   await t
>     .expect(header.languageSelector.exists)
>     .ok()
>     .click(header.languageSelector)
>     .click(header.languageSelectorItemSv)
>     .expect(getPathname())
>     .eql('/sv');
> ```

Github actually shows the stop in different place and in here it's related to the new headless cms feature. The path is not /sv but /sv/cms-page. Some reason why the cms-page is selected as a default page.

PR #147, opened 2021-09-06
PT-1138 | CMS navigation and SEO stuff

Sartonon:
## Description :sparkles:

- CMS navigation
- CMS pages SEO

### Closes :no_good_woman:

**[PT-1138](https://helsinkisolutionoffice.atlassian.net/browse/PT-1138):**


src/common/components/meta/PageMeta.tsx
```diff
@@ -9,22 +12,40 @@ export type Props = {
...
+    path: string;
+    locale: Language;
+  }[];
 };
 
+export function getLanguageAwarePath(
+  locale: string | undefined,
+  path: string
+): string {
+  if (locale === 'fi') {
```
nikomakela, line 30: Should this be taken from Language enum, so the places where the languages are used and checked could be found more easily?

src/headless-cms/components/CmsPageNavigation.tsx
```diff
@@ -1,52 +1,56 @@
...
+                            className={classNames({
+                              [styles.activeLink]: isActiveLink(uri),
+                            })}
+                          >
+                            <span>{n.title}</span>
+                          </a>
+                        </Link>
+                      );
+                    })}
+                </div>
```
nikomakela, line 47: This could be splitted to a new component so it would not look so clumsy

nikomakela (approved):

src/headless-cms/components/CmsPageNavigation.tsx
Sartonon: Which part? :D this is already a new component

PR #149, opened 2021-09-07
fix: language change when headless cms is not enabled

Sartonon:
### Closes :no_good_woman:


src/domain/app/header/Header.tsx
```diff
@@ -118,10 +118,16 @@ const Header: React.FC = () => {
...
                 lang={option.value}
                 label={option.label}
-                onClick={changeLanguage(
-                  `/cms-page${nav?.uri ? stripLocaleFormUri(nav?.uri) : ''}`,
-                  option.value
-                )}
+                onClick={
+                  isFeatureEnabled('HEADLESS_CMS')
+                    ? changeLanguage(
+                        `/cms-page${
```
nikomakela, line 124: A router constant should be used for path.

nikomakela (changes requested):

nikomakela (approved):

PR #150, opened 2021-09-08
feat(search): search by a division.

nikomakela:
PT-1130 PT-1168 PT-1169.

**Note: This feature needs the neighborhood API from Kultus API server!**

image


Sartonon (approved):
Lookin good :)

PR #155, opened 2021-09-17
feat: use allOngoingAnd parameter to get better search results

Sartonon:
## Description :sparkles:

- use allOngoingAnd parameter to get better search results

### Closes :no_good_woman:

**[PT-1206](https://helsinkisolutionoffice.atlassian.net/browse/PT-1206):**

### DIFF
--- a/src/domain/events/__tests__/EventsPage.test.tsx
+++ b/src/domain/events/__tests__/EventsPage.test.tsx
@@ -61,7 +61,7 @@ const mocks: MockedResponse[] = [
       variables: {
         include: ['keywords', 'location'],
         keyword: [],
-        text: '',
+        allOngoingAnd: null,
         inLanguage: '',
         location: '',
         start: 'now',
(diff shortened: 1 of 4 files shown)


nikomakela (commented):
Tested in my local environment. I rebuilt the cache with LE's management command and then the search seems to work ok when using keyword in free text field, but there are serious issues with an events counter and pagination. It may be so, because I'm using a local linked events and it might respond with wrong events (maybe the local cache is not in sync even after management command was ran). The page without filters was clear of events, which shouldn't be the case, but had a load more -button that rendered the event that is shown in the screenshot.

image


When tested with LE testserver, the page without filters loaded properly, but the free text search does not find any events when using keywords there. Maybe the staging server cache is out of sync?

So, locally, the keyword search works, but pagination is wrecked. With LE testserver, the pagination seems to work ok, but the search does not.

PR #156, opened 2021-09-17
fix: dont show upcoming occurrences in event card when there is none

Sartonon:
## Description :sparkles:

- Dont show upcoming occurrences in event card when there is none
- Show "No upcoming occurrences" text when event has no more upcoming occurrences

### Closes :no_good_woman:

**[PT-1208](https://helsinkisolutionoffice.atlassian.net/browse/PT-1208):**
**[PT-1164](https://helsinkisolutionoffice.atlassian.net/browse/PT-1164):**

### DIFF
--- a/src/domain/events/EventsPage.tsx
+++ b/src/domain/events/EventsPage.tsx
@@ -19,7 +19,6 @@ import {
   getEventFilterVariables,
   getInitialValues,
   getTextFromDict,
-  // getEventsThatHaveUpcomingOccurrence,
   getSearchQueryObject,
 } from './utils';
 
@@ -60,7 +59,6 @@ const EventsPage = (): ReactElement => {
   }, [router.query, organisationName]);
 
   const eventsWithUpcomingOccurrences = eventsData?.events?.data;
-  // getEventsThatHaveUpcomingOccurrence(eventsData);
 
   const search = (values: EventSearchFormValues) => {
     values = { ...values, organisation: values.organisationId };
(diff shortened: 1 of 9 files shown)


src/domain/events/EventsPage.tsx
```diff
@@ -59,8 +59,8 @@ const EventsPage = (): ReactElement => {
     };
   }, [router.query, organisationName]);
 
-  const eventsWithUpcomingOccurrences =
-    getEventsThatHaveUpcomingOccurrence(eventsData);
+  const eventsWithUpcomingOccurrences = eventsData?.events?.data;
+  // getEventsThatHaveUpcomingOccurrence(eventsData);
```
Sartonon, line 63: I still need to delete this commented line

nikomakela (changes requested):
Tested locally:

image

It says "no upcoming occurrences", but still renders the "other event times" -button.

Response from graphQl API:
```
pEvent: {id: "UGFsdmVsdXRhcmpvdGluRXZlbnROb2RlOjM3", nextOccurrenceDatetime: null,…}
id: "UGFsdmVsdXRhcmpvdGluRXZlbnROb2RlOjM3"
lastOccurrenceDatetime: "2021-04-30T22:30:00+00:00"
nextOccurrenceDatetime: null
```

The code checks 

```
const hasMultipleOccurrences = nextOccurrenceTime !== lastOccurrenceTime;
```

but the lastOccurrenceTime is given, even when it's in past.

The variable name should be changed and the button should be shown only when the last occurrence is in future.

nikomakela:
needless imports reported by automation tools could also be resolved when a requested change is full filled

nikomakela (approved):

PR #166, opened 2021-11-01
PT-1269 | New event page base layout

Sartonon:
## Description :sparkles:

First I updated next to version 12 if it would brind some perf gains in development. Changed couple js files to msj to get things working.

In summary:

- Update next to v12
- Implement new base layout for event page

### Closes :no_good_woman:

**[PT-1269](https://helsinkisolutionoffice.atlassian.net/browse/PT-1269):**

## Screenshots :camera_flash:

Screenshot 2021-11-01 at 15 50 37


nikomakela (commented):
I think this is very much approvable from a code perspective and the event page is looking nice. I'm not sure how much is achieved here yet and is this really a reusable base layout, but at least there is a new event hero! Looking good anyway 😄 .

nikomakela (approved):

Sartonon:
> I think this is very much approvable from a code perspective and the event page is looking nice. I'm not sure how much is achieved here yet and is this really a reusable base layout, but at least there is a new event hero! Looking good anyway 😄 .

My goal wasn't to create reusabled layout component yet because we realized that at this point there is only this page that will use it

PR #167, opened 2021-11-02
PT-1270 | Keywords as links

Sartonon:
## Description :sparkles:

Changed keyword sto be links to search page with keyword query parameter. Trickier part was to find out which keyword belongs to which keyword set. It made things little bit harder to implement :)

For example the following keywords have different keyword sets:

Kirjastovierailu -> `?additionalCriteria=kultus%3A20`
Satutunnit -> `?categories=yso%3Ap14710`
3-5 vuotiaat -> `?targetGroups=kultus%3A45`

In summary:

- Change keywords to links
- Add also audiences to keyword lists

## Issues :bug:

There is still issue with event cards because the card itself is a link and the keywords inside it are links too. Not quite sure yet what would be the best way to avoid this accessibility wise.

```
Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.
```

### Closes :no_good_woman:

**[PT-1270](https://helsinkisolutionoffice.atlassian.net/browse/PT-1270):**

### Automated tests :gear:️

- Added automated component tests to KeywordsList

## Screenshots :camera_flash:

Screenshot 2021-11-02 at 16 37 44

Screenshot 2021-11-02 at 16 37 37


nikomakela (approved):
Great job!

PR #168, opened 2021-11-04
PT-1269, PT-1271 | New search form panel and event search page

Sartonon:
## Description :sparkles:

- Add new search panel with different states (compact, condensed, advanced) <- hard to come up with good names or maybe there is another clearer way to do this different states :D
- Add new events search page

### Closes :no_good_woman:

**[PT-1269](https://helsinkisolutionoffice.atlassian.net/browse/PT-1269):**

## Screenshots :camera_flash:

Screenshot 2021-11-04 at 11 03 15
Screenshot 2021-11-04 at 11 03 11
Screenshot 2021-11-04 at 11 02 58
Screenshot 2021-11-04 at 11 02 52


public/locales/sv/events.json
```diff
@@ -30,6 +33,9 @@
     "labelTargetGroups": "Målgrupper",
     "labelActivities": "Aktiviteter",
     "placeInputPlaceholder": "Skriv sökord",
+    "buttonEditSearch": "SV: Edit search",
```
nikomakela, line 36: These needs translations? Are they asked from PO?

src/domain/events/__tests__/EventsSearchPage.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+      },
+    },
+  },
+];
+
+afterEach(() => {
+  clear();
+});
+
+test('renders search form and events list with correct information', async () => {
```
nikomakela, line 113: I think the focusing should be tested, because it is quite important in sense of accessibility. Another thing that could have tests is the expanded search when there are some filters active.

nikomakela (approved):
Translations should be requested and added, because the service is in the edge of heavy publishment :)

src/domain/events/__tests__/EventsSearchPage.test.tsx
nikomakela: I manually tested the focusing and tried the voice over also. The voice over speaks meaningful things, but I'm not sure if it's enough :)

public/locales/sv/events.json
Sartonon: yes

PR #173, opened 2021-11-11
feat: unit place selector uses schoolsAndKindergartens query

nikomakela:
API returns all the units (1000+ pcs) all at once from schoolsAndKindergartens graphql query. The Enrolment form autosuggest field for unit selection should be able to filter the list of units with an input value.

PT-1314.

----
The list of schools and kindergartens is from Servicemap API, but the selected place is queried from LinkedEvents places-API

Selected:
image

With error:
image

### DIFF
--- a/src/domain/place/placeText/PlaceText.tsx
+++ b/src/domain/place/placeText/PlaceText.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { useTranslation } from 'react-i18next';
 
 import { usePlaceQuery } from '../../../generated/graphql';
 import useLocale from '../../../hooks/useLocale';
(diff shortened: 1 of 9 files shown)


nikomakela:
- [ ] Tests are still missing

src/domain/place/placeText/PlaceText.tsx
```diff
@@ -1,20 +1,26 @@
...
+  const { t } = useTranslation();
   const { data } = usePlaceQuery({
     variables: { id: placeId },
   });
+  const text = getLocalisedString(data?.place?.name || {}, locale);
+  const errorText =
+    showErrorText && placeId && !text ? t('placeText.noPlaceFoundError') : '';
 
-  return <>{getLocalisedString(data?.place?.name || {}, locale)}</>;
+  return <>{text || errorText}</>;
```
Sartonon, line 23: This feels unnecessarily complicated. Maybe could get rid of `!text` because it is redundant with this code.

Sartonon: I also think that `t('placeText.noPlaceFoundError')` shouldn't be in this component but rather provided as errorText prop

Sartonon: Because this is general reusable component and text "Paikka on valittu..." is specific to UnitPlaceSelector.

Sartonon (changes requested):
Otherwise looks good but some code should be refactored out of PlaceText component :)

nikomakela:
The acceptance tests are failing again 😞 . I'm quite sure it's related to a Linkedevent cache.

PR #177, opened 2021-11-19
feat: event back button navigates to correct page

Sartonon:
## Description :sparkles:

- Event page back button now navigates to correct page preserving query parameters

### Closes :no_good_woman:

**[PT-1293](https://helsinkisolutionoffice.atlassian.net/browse/PT-1293):**


nikomakela (approved):
Our routing is so complex that it feels like it should not work, but I tested that it seems to work ok! 😄 Well done!

PR #178, opened 2021-11-21
PT-1273 |  Use new upcoming events query in front page

Sartonon:
### Closes :no_good_woman:

### DIFF
--- a/src/domain/events/EventsPage.tsx
+++ b/src/domain/events/EventsPage.tsx
@@ -2,20 +2,28 @@ import { Koros } from 'hds-react';
 import { useTranslation } from 'next-i18next';
 import Head from 'next/head';
 import Link from 'next/link';
+import { useRouter } from 'next/router';
 import React, { ReactElement } from 'react';
 
 import LoadingSpinner from '../../common/components/loadingSpinner/LoadingSpinner';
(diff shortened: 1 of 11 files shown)


nikomakela:
Don't know if it really matters, but to note, the occurrences are in "strange order" when the next occurrence is the first but it's not the first one. Maybe the occurrences in past could be hidden?
image

nikomakela:
The title could be "The upcoming events" or "The next upcoming events"

image

src/domain/events/EventsPage.tsx
```diff
@@ -2,43 +2,87 @@ import { Koros } from 'hds-react';
...
+          variables: {
+            page: nextPage,
+          },
+        });
+        setIsLoadingMore(false);
+      } catch (e) {
+        setIsLoadingMore(false);
+      }
+    }
+  };
```
nikomakela, line 85: I don't think the upcoming events will ever have a pagination, because if paginated in Kultus, The pagination from LinkedEvents would be a problem. I may be wrong, but I thought the API implementation is so that the next is always `null`. Sure it does not wreck anything and UI doesn't need to know such a detail 😄 

How is it @charn?

nikomakela (approved):
There is still something to consider about and I added some comments about them, but the code looks good to me and the overall is approvable.

PR #182, opened 2021-11-25
fix: unit id fields entries are filtered in lowercase format

nikomakela:
PT-1343

image

### DIFF
--- a/src/domain/place/unitSelector/UnitPlaceSelector.tsx
+++ b/src/domain/place/unitSelector/UnitPlaceSelector.tsx
@@ -74,7 +74,9 @@ const UnitPlaceSelector: React.FC<Props> = ({
         value: place.id || '',
       }))
       // Filter the results with a search value!
-      .filter((place) => place.label.includes(searchValue)) || [];
+      .filter((place) =>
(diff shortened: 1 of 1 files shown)


src/domain/place/unitSelector/UnitPlaceSelector.tsx
```diff
@@ -74,7 +74,9 @@ const UnitPlaceSelector: React.FC<Props> = ({
         value: place.id || '',
       }))
       // Filter the results with a search value!
-      .filter((place) => place.label.includes(searchValue)) || [];
+      .filter((place) =>
+        place.label.toLowerCase().includes(searchValue.toLowerCase())
+      ) || [];
```
nikomakela, line 79: I also tried a really lightweight plugin that uses `localeCompare()` with `includes()`. It worked, but the field and the input became really laggy, so I decided to go with really fast and simple `toLowerCase()`.
```suggestion
      # import { localeIncludes } from 'locale-includes';
      .filter((place) =>
        localeIncludes(place.label, searchValue, {
          usage: 'search',
          sensitivity: 'base',
        })
      ) || [];
```

Sartonon (approved):
Looks good :)

PR #184, opened 2021-11-26
PT-1316 | Add language column to occurrences table in event page

Sartonon:
## Description :sparkles:

- Add language column to occurrences table in event page

### Closes :no_good_woman:

**[PT-1316](https://helsinkisolutionoffice.atlassian.net/browse/PT-1316):**

## Screenshots :camera_flash:
Screenshot 2021-11-26 at 9 35 59

### DIFF
--- a/src/domain/event/occurrences/OccurrencesTable.tsx
+++ b/src/domain/event/occurrences/OccurrencesTable.tsx
@@ -216,6 +216,12 @@ const OccurrenceEnrolmentTable: React.FC<{
       id: 'time',
     },
     {
+      Header: t('enrolment:occurrenceTable.columnLanguage'),
+      accessor: (row: OccurrenceFieldsFragment) =>
+        row.languages.edges.map((lang) => lang?.node?.id).join(', ') ?? '-',
+      id: 'languages',
+    },
+    {
       Header: t('enrolment:occurrenceTable.columnPlace'),
       accessor: (row: OccurrenceFieldsFragment) => {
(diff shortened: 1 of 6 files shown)


src/domain/event/occurrences/OccurrencesTable.tsx
```diff
@@ -215,6 +215,12 @@ const OccurrenceEnrolmentTable: React.FC<{
 
       id: 'time',
     },
+    {
+      Header: t('enrolment:occurrenceTable.columnLanguage'),
+      accessor: (row: OccurrenceFieldsFragment) =>
+        row.languages.edges.map((lang) => lang?.node?.id).join(', ') ?? '-',
+      id: 'languages',
+    },
```
nikomakela, line 223: aria-label should interpret the language in full, not as abbreviations. It was also mentioned in the Jira-ticket.

nikomakela (changes requested):

src/domain/event/occurrences/OccurrencesTable.tsx
nikomakela: Actually, when tested with a voice over, all the show details buttons are the same and the row is not interpreted at all. I think the row needs to be interpreted by a voice over somehow. Maybe the row content should be included in the buttons aria-label?

nikomakela: I can approve this PR without accessibility changes, but an accessibility ticket needs to be created then.

PR #188, opened 2021-11-30
PT-1358 | Disallow enrolling when it has not begun

Jokinen:
## Description :sparkles:

Hide enrolment button until enrolling has begun.

Instead show text that describes when enrolment will begin.

Also contains:

* Handles error states within additional information

### Closes :no_good_woman:

**[PT-1358](https://helsinkisolutionoffice.atlassian.net/browse/PT-1358):**

### Automated tests :gear:️

Added a test case to check that the button does not exist if enrolment has not begun. A lot of tests already so I was a bit hesitant whether I should add to the maintenance load. In the end I thought that this behaviour is worth testing as it has a big enough impact on UX and could be easy to break because there's a lot of boolean logic where this button is declared. Can be dropped if not.

## Screenshots :camera_flash:

Screenshot 2021-11-30 at 13 30 34

### DIFF
--- a/src/domain/event/__tests__/EventPage.test.tsx
+++ b/src/domain/event/__tests__/EventPage.test.tsx
@@ -560,8 +560,20 @@ it('opens expanded area with enrolment button when clicked', async () => {
 });
 
 it('expanded area does not have an enrolment button if enrollment has not yet begun', async () => {
-  advanceTo(new Date(2020, 6, 12)); // Before enrolmentStart (2020-07-13T06:00:00+00:00)
-  renderComponent();
+  advanceTo(new Date(2020, 6, 10)); // Before enrolmentStart (2020-07-13T06:00:00+00:00)
+  renderComponent({
+    mocks: [
+      createEventQueryMockIncludeLanguageAndAudience({
+        ...eventData,
+        id: eventData.id as string,
+        pEvent: fakePEvent({
+          ...eventData.pEvent,
+          neededOccurrences: 1,
+        }),
+      }),
+      ...apolloMocks.slice(1),
+    ],
+  });
   await waitForRequestsToComplete();
 
   const occurrenceRow = within(screen.getAllByRole('row')[8]);
--- a/src/domain/event/__tests__/EventPage.test.tsx
+++ b/src/domain/event/__tests__/EventPage.test.tsx
@@ -581,7 +581,7 @@ it('expanded area does not have an enrolment button if enrollment has not yet be
       name: 'Ilmoittaudu',
     })
   ).toEqual(null);
-  expect(screen.queryByText(/Ilmoittautuminen aukeaa/)).toBeInTheDocument();
+  expect(screen.queryByText(/Ilmoittautuminen alkaa/)).toBeInTheDocument();
 });
 
 it('filters occurrence list correctly when sate filters are selected', async () => {
--- a/src/domain/event/__tests__/EventPage.test.tsx
+++ b/src/domain/event/__tests__/EventPage.test.tsx
@@ -224,8 +224,6 @@ const apolloMocks = [
   ]),
 ];
 
-advanceTo(new Date(2020, 6, 14));
-
 const renderComponent = (options?: CustomRenderOptions) => {
   return render(<EventPage />, {
     mocks: apolloMocks,
@@ -250,6 +248,10 @@ beforeAll(() => {
 
 afterAll(() => {
   jest.setTimeout(5000);
+});
+
+beforeEach(() => {
+  advanceTo(new Date(2020, 6, 14));
 });
 
 it('shows full events correctly in occurrences list', async () => {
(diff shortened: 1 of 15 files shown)


src/domain/event/occurrences/OccurrenceInfo.tsx
```diff
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import { isSameDay } from 'date-fns';
+import { isSameDay, isAfter } from 'date-fns';
```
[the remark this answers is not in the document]
Sartonon: Tried to come up with working rule to avoid these import in the future but couln't get it working with no-restricted-imports eslint rule

src/domain/event/occurrences/OccurrenceInfo.tsx
```diff
@@ -57,6 +58,8 @@ const OccurrenceInfo: React.FC<{
   const autoAcceptance = event.pEvent?.autoAcceptance;
   const externalEnrolmentUrl = event.pEvent.externalEnrolmentUrl ?? '';
   const hasExternalEnrolment = !!event.pEvent.externalEnrolmentUrl;
+  const enrolmentStart = event.pEvent.enrolmentStart;
+  const isEnrollmentOpen = isAfter(new Date(enrolmentStart), new Date());
```
Sartonon, line 62: Should be `isEnrolmentOpen` with one `l` to keep naming consistent :)

src/domain/event/occurrences/OccurrenceInfo.tsx
```diff
@@ -181,14 +184,15 @@ const OccurrenceInfo: React.FC<{
         language={locale}
         variant="button"
       />
-      {hasExternalEnrolment ? (
+      {hasExternalEnrolment && (
```
Sartonon, line 187: Seems like you got rid of nested ternary which is fine for me but I think only one of these buttons or only the enrolment start notice should be rendered at a time 🤔

src/domain/event/occurrences/OccurrenceInfo.tsx
Sartonon: Also `occurrence/utils` already has util to check if enrolment is open:

```
 isEnrolmentClosed,
  hasOccurrenceSpace,
  isOccurrenceCancelled,
  isEnrolmentStarted,
```

src/domain/event/occurrences/OccurrenceInfo.tsx
Sartonon: This might also need checks if enrolment has ended. We had that previously but seems like I forgot to add those there during the redesign.

Seems like old component `EnrolmentButtonCell` is still in the project. You could take a look how enrolment button and other notices were rendered previously.

src/domain/event/__tests__/EventPage.test.tsx
```diff
@@ -527,6 +529,31 @@ it('opens expanded area with enrolment button when clicked', async () => {
...
+
+  await waitFor(() => {
+    expect(screen.queryByText(data.venueDescription)).toBeInTheDocument();
+  });
+
+  expect(
+    screen.queryByRole('button', {
+      name: 'Ilmoittaudu',
+    })
+  ).toEqual(null);
```
Sartonon, line 553: `).not.toBeInTheDocument();` would be consistent with how these checks are done in other tests

Sartonon: `).not.toBeInTheDocument();` might even give better error message in console if this assertion fails

Sartonon (changes requested):
We might need to do same checks and render same notices as EnrolmentButtonCell did previously. Seems like I had totally forgotten those during EventPage redesign.

src/domain/event/__tests__/EventPage.test.tsx
Jokinen: I had this in the first version, but refactored it later. Seems more precise to check for `null` because it's a value that doesn't really relate with the DOM. But I can change this.

Sartonon: Of course this would maybe be the most readable way to do this. But I understand your point of view
```
const enrolmentButton = screen.queryByRole('button', {
      name: 'Ilmoittaudu',
 });
 expect(enrolmentButton).not.toBeInTheDocument();
```

Sartonon: But I think most importantly `toBeInTheDocument` should be used to stay consistent in this project

src/domain/event/occurrences/OccurrenceInfo.tsx
Jokinen: > Seems like you got rid of nested ternary which is fine for me but I think only one of these buttons or only the enrolment start notice should be rendered at a time 🤔

Doesn't it work in this way in this changeset? Buttons and notices are not shown at the same time.

Do you mean that external enrolments also require a notice?

Sartonon: I think it does work correctly but having `!hasExternalEnrolment && neededOccurrences === 1 && !isEnrollmentOpen` and `hasExternalEnrolment` as separate checks makes this part of the code hard to understand

Jokinen: Ah, now I maybe understood you. Do you have something in mind to fix it with?

This part of the logic in the component is hard to reason about, but I feel like these changes haven't increased or decreased it by much. Not chaining elvis is something people consistently recommend so it seemed like a good low hanging fruit to pick.

The next things I'd be looking at is to formalise the rendering in some way--perhaps by having a state string that describes the button that should be rendered. But imo that's a bit heavy for now.

Now that we are handling the errors in this PR as well, this part may need reafctoring.

Sartonon: I think the easiest way to make this easier to reason about would be to have function/Component that just has if statements with early returns like `EnrolmentButtonCell.tsx` did previously. That way it is pretty clear that only one thing is rendered at a time

Sartonon: Which would be some kind of notice or enrolment button of 2 different kinds

Jokinen: This got refactored while I added the error handling.

Sartonon (approved):
Looks very good :)

PR #191, opened 2021-12-02
PT-1347 | Fix enrolment form group size validation bug

Sartonon:
## Description :sparkles:

Persisting min and maxGroupSize values that were only used for validation was causing problems because previous values would be taken from local storage. To fix this min and MaxGroupSize values are now passed ar parameters to getValidationSchema to avoid persisting them to local storage.

### Closes :no_good_woman:

**[PT-1347](https://helsinkisolutionoffice.atlassian.net/browse/PT-1347):**

### DIFF
--- a/src/domain/event/occurrences/OccurrenceInfo.tsx
+++ b/src/domain/event/occurrences/OccurrenceInfo.tsx
@@ -1,3 +1,4 @@
+import { useApolloClient } from '@apollo/client';
 import classNames from 'classnames';
 import { isSameDay } from 'date-fns';
 import {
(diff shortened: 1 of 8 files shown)


src/domain/event/occurrences/OccurrenceInfo.tsx
```diff
@@ -75,6 +78,13 @@ const OccurrenceInfo: React.FC<{
     enrolButtonRef.current?.focus();
   };
 
+  const handleOnEnrol = async () => {
+    // refetch event query for the page to get updated occurrences
+    await apolloClient.refetchQueries({
+      include: [EventDocument],
+    });
+  };
```
Sartonon, line 86: Added also refetch here to get updated seats after enrolling.

nikomakela (approved):
The code changes looks good to me, but I'm not sure how to test this. I don't want to delay the deployment, so I can approve this :P

PR #193, opened 2021-12-03
fix: change language order in occurrence table to be fi, sv first

Sartonon:
## Description :sparkles:

- Update some test to use inline snapshot
- Fix issue with languages that have underscore in their ID
- Order languages correctly (fi, sv...)

### Closes :no_good_woman:

**[PT-1316](https://helsinkisolutionoffice.atlassian.net/browse/PT-1316):**


nikomakela (approved):
looks good

PR #194, opened 2021-12-05
PT-1337 | Use local storage to remember if user close notification

Sartonon:
## Description :sparkles:

Notification hash and closed state are persisted to local storage. Notification is visible to the user again if any of the notification fields are changed in the CMS. Hash that is generated from the notification fields is saved to local storage and always compared with the one fetched from the CMS

### Closes :no_good_woman:

**[PT-1337](https://helsinkisolutionoffice.atlassian.net/browse/PT-1337):**

### DIFF
--- a/src/utils/hash.ts
+++ b/src/utils/hash.ts
@@ -1,0 +1,20 @@
+// https://stackoverflow.com/a/52171480
+export default function hash(str: string, seed = 0): number {
+  let h1 = 0xdeadbeef ^ seed,
+    h2 = 0x41c6ce57 ^ seed;
+
+  for (let i = 0, ch: number; i < str.length; i = i + 1) {
+    ch = str.charCodeAt(i);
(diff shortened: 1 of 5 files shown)


src/utils/hash.ts
```diff
@@ -0,0 +1,20 @@
+// https://stackoverflow.com/a/52171480
+export default function hash(str: string, seed = 0): number {
```
nikomakela, line 2: I find this a bit exaggerated, but it definitely works and gives an unique hash :) Something far simpler would have been enough in this case, but its ok to use this ofc

nikomakela (approved):
Tested (with phone) and its working well

src/utils/hash.ts
Sartonon: Copy pasted from liikunta-project :)

PR #195, opened 2021-12-07
fix: cms card layout

Sartonon:
## Issues :bug:

- Fix cms cards :)

### Closes :no_good_woman:

**[PT-1396](https://helsinkisolutionoffice.atlassian.net/browse/PT-1396):**

## Screenshots :camera_flash:
Screenshot 2021-12-07 at 11 29 14


nikomakela (approved):
Acceptable! :) Better than old which exploited the image when there was no lead text.

PR #197, opened 2021-12-07
PT-1366 | Allow user to load more upcoming events

Jokinen:
## Description :sparkles:

Adds pagination to the list of upcoming events on the landing page.

Allows users to load more upcoming events.

Uses existing pagination logic in `EventList`.

### Closes :no_good_woman:

**[PT-1366](https://helsinkisolutionoffice.atlassian.net/browse/PT-1366):**

### Automated tests :gear:️

* Added check for existence of load more events button

## Screenshots :camera_flash:

Screenshot 2021-12-07 at 18 00 47


src/domain/events/EventsPage.tsx
```diff
@@ -160,5 +169,48 @@ const EventsPageMeta: React.FC = () => {
...
+  });
+  const events = eventsData?.upcomingEvents?.data;
+  const eventsCount = eventsData?.upcomingEvents?.pageInfo.totalCount;
+  const hasNextPage = eventsData?.upcomingEvents?.pageInfo.hasNextPage;
+  const page = eventsData?.upcomingEvents?.pageInfo.page ?? 0;
+
+  const nextPage = React.useMemo(
+    () => (hasNextPage ? page + 1 : null),
+    [hasNextPage, page]
+  );
```
Sartonon, line 194: I honestly think that `React.useMemo` is event less perfomant than without.

Sartonon (approved):
Looks good 👍  

I think I would just get rid of the useMemo :)

```
 const nextPage = React.useMemo(
    () => (hasNextPage ? page + 1 : null),
    [hasNextPage, page]
  );
```

src/domain/events/EventsPage.tsx
Jokinen: Good point 👍

PR #201, opened 2021-12-09
PT-1321 | Add support for email link for enrolment button

Sartonon:
## Description :sparkles:

- Add support for email link for enrolment button

### Closes :no_good_woman:

**[PT-1321](https://helsinkisolutionoffice.atlassian.net/browse/PT-1321):**

- Should open email client when clicking enrolment link:

test url: http://localhost:3000/event/palvelutarjotin:af6z672j6m

### DIFF
--- a/src/common/components/externalLink/ExternalLink.tsx
+++ b/src/common/components/externalLink/ExternalLink.tsx
@@ -6,18 +6,12 @@ import SrOnly from '../SrOnly/SrOnly';
 
 type Props = React.AnchorHTMLAttributes<HTMLAnchorElement> & {
   href: string;
-  isEmail?: boolean;
 };
 
-const ExternalLink: React.FC<Props> = ({ children, isEmail, ...props }) => {
+const ExternalLink: React.FC<Props> = ({ children, ...props }) => {
   const { t } = useTranslation();
   return (
-    <a
-      {...props}
-      href={isEmail ? `mailto:${props.href}` : getExternalUrl(props.href)}
-      target="_blank"
-      rel="noreferrer"
-    >
+    <a {...props} href={props.href} target="_blank" rel="noreferrer">
       {children}
       <IconLinkExternal
         style={{ verticalAlign: 'middle', marginLeft: '0.5rem' }}
@@ -27,7 +21,7 @@ const ExternalLink: React.FC<Props> = ({ children, isEmail, ...props }) => {
   );
 };
 
-const getExternalUrl = (url: string) => {
+export const getExternalUrl = (url: string): string => {
   if (url.match(/^https?:\/\//)) {
     return url;
   }
--- a/src/common/components/externalLink/ExternalLink.tsx
+++ b/src/common/components/externalLink/ExternalLink.tsx
@@ -4,14 +4,17 @@ import * as React from 'react';
 
 import SrOnly from '../SrOnly/SrOnly';
 
-type Props = React.AnchorHTMLAttributes<HTMLAnchorElement> & { href: string };
+type Props = React.AnchorHTMLAttributes<HTMLAnchorElement> & {
+  href: string;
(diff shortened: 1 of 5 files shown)


src/domain/event/occurrences/EnrolmentButtonCell.tsx
```diff
@@ -8,6 +8,7 @@ import {
   OccurrenceFieldsFragment,
 } from '../../../generated/graphql';
 import useLocale from '../../../hooks/useLocale';
+import isEmail from '../../../utils/isEmail';
```
nikomakela, line 11: Unused import

nikomakela (approved):
Great!

Tested clicking it:
image

src/common/components/externalLink/ExternalLink.tsx
```diff
@@ -4,14 +4,17 @@ import * as React from 'react';
 
 import SrOnly from '../SrOnly/SrOnly';
 
-type Props = React.AnchorHTMLAttributes<HTMLAnchorElement> & { href: string };
+type Props = React.AnchorHTMLAttributes<HTMLAnchorElement> & {
+  href: string;
+  isEmail?: boolean;
```
Jokinen, line 9: For me `isEmail` is a bit specific when it comes to the API of this component. I assume that it's used here because of the `getExternalUrl` check. I would say that ideally this component does not know or care what valid `href` value it receives--I would expect that it would result in less code.

That being said, this is more of a FYI comment--this approach works for me if it's pragmatic etc.

Jokinen (approved):
This looks good to me. General concerns I have:

* If we think of data integrity, giving our support for storing `email` and `url` values in the same field may be something that later generations wonder about. AFAIK the naming of the field still specifies `url`. I'm just saying this aloud in case it hasn't been said yet. The distinction isn't huge here technically, but convention usually places a greater difference between `emails` and `urls`.
* By clicking the link, and email compose window is opened. I wonder, what should I as a user do next? Shouldn't the user get some instructions on what information they should provide etc? Was the idea that these would be included in the description of the event? Couldn't it be beneficial to the provider that the could set some default title that would make it easier to find enrolments later on?

I recognise that with this change were are trying to do a very small change in order to enable providers to use the application in the way they have attempted to and for that purpose these changes work for me. I wanted to highlight these points just to ensure that they have been voiced aloud at some point.

src/common/components/externalLink/ExternalLink.tsx
Sartonon: How would you change the API? Would you move `getExternalUrl` out of this component?

nikomakela: This is our field in a Django model:
```
    external_enrolment_url = models.URLField(
        verbose_name=_("enrolment url"), blank=True, null=True
    )
```
When the field is created, it will be a basic varchar-field
```
CREATE TABLE occurrences_palvelutarjotinevent (
    -- ...
    external_enrolment_url character varying(200)
);
```
So the URL-validation is purely on ORM-level and even there it needs to be manually called. @Jokinen is right that it is a bit bad for data integrity that 1 field stores both the web-addresses and the mail-links, but I don't see much data value in this field anyway. The thing is that this is very important feature for the product owner and the deadline is only a couple of days away. 

I would create another ticket where we could investigate other options, make changes in implementation and run a data migration for the data. If we don't publish this feature now, we might not get the data at all, because the final users won't use the system or the issue would be raised as a bug ticket immediately after the launch.

nikomakela: ...The data migration should be easy, since only all the emails contains "@", so we could find those rows and move the dat a from a field to another.

Sartonon: Refactored ExternalLink a little bit

PR #202, opened 2021-12-13
feat: newsletter subscription page

nikomakela:
PT-1122.

image


nikomakela:
- [ ] Tests are still missing

public/locales/fi/newsletter.json
```diff
@@ -0,0 +1,20 @@
...
+      "earlyChildhoodEducation": "Varhaiskasvatukselle",
+      "highAndSecondarySchools": "Yläkouluille ja toisen asteen oppilaitoksille"
+    },
+    "labelGroups": "Tilaan uutiskirjeen",
+    "labelFirstName": "Etunimi",
+    "labelLastName": "Sukunimi",
+    "labelEmail": "Sähköposti",
+    "buttonSubmit": "Tilaa uutiskirje",
+    "subscribeSuccess": "Tilaus lähetetty. Kiitos tilauksesta!",
+    "subscribeFailed": "subscribeFailed"
```
Sartonon, line 14: translation :)

nikomakela:
- [ ] Actually, a link to newsletter is still missing from footer

image

.github/workflows/review.yml
```diff
@@ -32,6 +32,8 @@ jobs:
           DOCKER_BUILD_ARG_NEXT_PUBLIC_API_BASE_URL: 'https://palvelutarjotin-api.test.kuva.hel.ninja/graphql'
           DOCKER_BUILD_ARG_NEXT_PUBLIC_UNIFIED_SEARCH_BASE_URL: 'https://unified-search.test.kuva.hel.ninja/search'
           DOCKER_BUILD_ARG_NEXT_PUBLIC_CMS_BASE_URL: 'https://kultus.content.api.hel.fi/graphql'
+          DOCKER_BUILD_ARG_NEWSLETTER_BASE_URL: 'https://api.createsend.com/api/v3.2'
+          DOCKER_BUILD_ARG_NEWSLETTER_APIKEY: ${{ secrets.GH_QA_NEWSLETTER_APIKEY }}
```
nikomakela, line 36: Should these be in env-section?

nikomakela: They are not public after the implementation was moved to next js api.

Sartonon: I think this is the correct place

Sartonon (commented):
Looks good. Couple comment added :)

Sartonon:
Also I think that form should reset after succesful subscription. And also maybe we should use HDS Notification component to inform user

src/pages/api/newsletter/subscribe/[groupId].ts
```diff
@@ -0,0 +1,54 @@
...
+      if (subscriberMail) res.json({ email: subscriberMail.data });
+      break;
+    case 'DELETE':
+      const response = await deleteSubscriber(groupId, email).catch((err) =>
+        axiosErrorHandler(res, err)
+      );
+      if (response) {
+        res.json({ email });
+      }
+      break;
```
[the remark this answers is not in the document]
Comment (nikomakela) [reply src/pages/api/newsletter/subscribe/[groupId].ts:47]: After thinking this a while, I think we actually need this feature (for subscription details deletion) in sake of law. I'm not sure about this and it wasn't in the scope of the original ticket, but I think it's the same kind of thing as cookie consents, etc.

I was already removing the code, but then reverted them back with some additional comments.

What do you think?

src/pages/api/newsletter/subscribe/[groupId].ts
```diff
@@ -0,0 +1,54 @@
...
+    ? query.groupId[0]
+    : query.groupId;
+  const email = Array.isArray(query.email) ? query.email[0] : query.email;
+  switch (method) {
+    case 'GET':
+      const subscriber = await getDetails(groupId, email).catch((err) =>
+        axiosErrorHandler(res, err)
+      );
+      if (subscriber) res.json(subscriber.data);
+      break;
```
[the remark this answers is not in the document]
Comment (nikomakela) [reply src/pages/api/newsletter/subscribe/[groupId].ts:33]: I think we might need this in very near future: see the comments under the delete -case.

src/pages/api/newsletter/subscribe/[groupId].ts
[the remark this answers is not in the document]
Comment (Sartonon) [reply src/pages/api/newsletter/subscribe/[groupId].ts]: No need to remove this at least for now

src/domain/newsletter/SubscribeNewsletterPage.tsx
```diff
@@ -0,0 +1,97 @@
...
+} from './newsletterSubscribeForm/NewsletterSubscribeForm';
+import styles from './subscribeNewsletterPage.module.scss';
+
+const SubscribeNewsletterPage: React.FC = () => {
+  const { t } = useTranslation();
+  const [notification, setNotification] = React.useState<JSX.Element | null>(
+    null
+  );
+  const [notificationAutoCloseOpen, setNotificationAutoCloseOpen] =
+    React.useState(false);
```
Sartonon, line 25: I think we shouldn't put react components to the state (it works but we could have simpler solution). I think we could have one state with `'success'` and `'error'` states, `null` would mean that notification is closed. Then render Notificarion based on the state

Sartonon (approved):
Amazing ⭐

PR #207, opened 2021-12-15
feat: add support for localization in notification

Sartonon:
## Description :sparkles:

- Added language parameter to Notification query

### Closes :no_good_woman:


nikomakela (commented):
The code changes looks good and The language selection really changes the notification text, but e.g Swedish text is not working correctly. The issue might be in Headless CMS -end, but the problem really exists:

image

I also think that we are using the notification wrong from headless CMS admin-ui. I think the text should always be changed to "Asetukset" -page and there should not be any other pages.

nikomakela (approved):
Now it seems to work ok when there is only 1 "asetukset" -configuration, so I'll approve this! Nice work!

PR #210, opened 2021-12-23
new dockerfile

nikomakela:
PT-1122


nikomakela:
Was used only for testing...

PR #212, opened 2022-01-07
feat(newsletter): show privacy statement in the newsletter form

nikomakela:
PT-1425 PT-1427

image


src/domain/newsletter/SubscribeNewsletterPage.tsx
```diff
@@ -75,9 +77,13 @@ const SubscribeNewsletterPage: React.FC = () => {
...
-          <TextWithLineBreaks
-            as="p"
-            text={t('newsletter:subscribeNewsletterPage.leadText')}
+          <p
+            dangerouslySetInnerHTML={{
+              __html: t('newsletter:subscribeNewsletterPage.leadText', {
+                openInNewTab: t('common:srOnly.opensInANewTab'),
+                url: PRIVACY_POLICY_LINKS[locale],
+              }),
+            }}
```
Jokinen, line 86: You might be able to avoid `dangerouslySetInnerHTML` and defining HTML in the translation data by using the `Trans` component:

https://react.i18next.com/latest/trans-component#alternative-usage-which-lists-the-components-v11.6.0

Jokinen (commented):
The link has a cue for screen readers, but is missing a visual cue that communicates that it'll open in a new window for visual users.

https://www.w3.org/TR/WCAG20-TECHS/H83.html

>For those not using assistive technology, the indication would also be available from the link text.

Icon etc. is sufficient. First recommendation is to open the window in place instead of a new window.

Sartonon (approved):
Looks good

Jokinen (approved):

PR #213, opened 2022-01-11
feat(newsletter): delete Gruppo newsletter account button

nikomakela:
PT-1424 PT-1444. Allow the delete method to Gruppo newsletter API.


nikomakela:
The PR is left as draft, because I'm not happy yet with the design related to the account deletion.

PR #216, opened 2022-01-18
PT-1451 | Add sidebar layout

Jokinen:
## Description :sparkles:

Adds support for sidebar layout for CMS pages when sidebar has content.

Retains old single column layout for pages that do not have any sidebar content.

### Closes :no_good_woman:

**[PT-1451](https://helsinkisolutionoffice.atlassian.net/browse/PT-1451)**

### Automated tests :gear:️

Added a test case into the integration tests of the CMS page that checks that the sidebar layout is used when the sidebar has content. This case can later be extended to check that sidebar content is rendered correctly.

This was a bit hazardous in that the test cases can't be run, because the bootstrapping library that's used does not support our version of next.

Used `/cms-page/kulttuurikasvatus` to verify behaviour with and without sidebar content.

## Screenshots :camera_flash:

Screenshot 2022-01-18 at 9 50 14
Screenshot 2022-01-18 at 9 50 37

*Sidebar comes after main content on mobile*


src/headless-cms/components/CmsPageContent.tsx
```diff
@@ -1,25 +1,19 @@
...
-          <HtmlToReact>{content}</HtmlToReact>
-        </>
-      )}
-    </Container>
-  );
+  if (sidebarContents.length > 0) {
+    return <CmsSidebarLayout page={page} />;
+  }
+
+  return <CmsOneColumnLayout page={page} />;
```
Sartonon, line 16: Was the idea for the content to always be 70% wide? Even when sidebar is missing?

Jokinen: Oh, you are correct. I enabled sidebar layout for all views.

src/headless-cms/components/cmsSidebarLayout.module.scss
```diff
@@ -0,0 +1,22 @@
...
+      flex-basis: 66.666%;
+    }
+  }
+
+  &Aside {
+    @include layout.respond-above(md) {
+      flex-basis: 33.333%;
+    }
+  }
+}
```
Sartonon, line 22: Hmm. I'm surprised that you chose to use flexbox when styling this layout 🤔 Did you consider grid?

My solution would be 

```
.cmsSidebarLayout {
  display: grid;

  @include layout.respond-above(md) {
    grid-template-columns: 2fr 1fr;
  }
}

```

src/headless-cms/components/CmsPageContent.tsx
```diff
@@ -1,25 +1,12 @@
...
-    <Container>
-      {page && (
-        <>
-          <h1>{title}</h1>
-          <HtmlToReact>{content}</HtmlToReact>
-        </>
-      )}
-    </Container>
-  );
+  return <CmsSidebarLayout page={page} />;
```
Sartonon, line 9: Are you planning to leave this here? Do we need CmsSidebarLayout as a separate component?

src/__tests__/cms-page/[...slug].test.tsx
```diff
@@ -133,6 +141,24 @@ describe.skip('CMS Page', () => {
     ).toBeInTheDocument();
     expect(screen.queryByText(page.content)).toBeInTheDocument();
   });
+
+  it('renders with sidebar layout when sidebar has content', async () => {
```
Comment (Sartonon) [inline src/__tests__/cms-page/[...slug].test.tsx:145]: I think these integration tests should be done for CmsPage.tsx component now that our testing library is not working.

src/headless-cms/components/cmsSidebarLayout.module.scss
Jokinen: I used flex because this is a one dimensional layout and in general the guideline tends to be to use flex with 1D and grid with 2D.

src/headless-cms/components/CmsPageContent.tsx
Jokinen: I was planning to remove it in a later PR to avoid a merge conflict.

Sartonon (approved):

PR #217, opened 2022-01-18
PT-1449 | Add support for layout link list

Jokinen:
## Description :sparkles:

Add support for layout link list in CMS sidebar.

`backgroundColor` is not supported. I consulted the designer about what impact it should have. We checked https://nuorten.helsinki/ for examples, but as far as  could tell, they hadn't made use of this setting.

The admin UI promotes the use of URLs that link against the API. We discussed this briefly in Slack and one suggestion was that the default behaviour can be configured to return URLs that link to clients built with this codebase. This change has not been made yet.

### Closes :no_good_woman:

**[PT-1449](https://helsinkisolutionoffice.atlassian.net/browse/PT-1449)**

### Automated tests :gear:️

Extended existing test case. However, wasn't able to run these tests. I wonder if it'd be worth it to transition into using some other way to bootstrap the tests. The linked GH issue gives the impression that the problems won't be fixed in a while.

This content has been configured with a link list: `en/cms-page/kulttuurikasvatus`

## Screenshots :camera_flash:

Screenshot 2022-01-18 at 17 04 14
Screenshot 2022-01-18 at 17 04 53

According to the designer, the design should mimic that of nuorten.helsinki. At places I have adopted visual approaches that aren't visible in the design, but can be seen in nuorten.helsinki.


src/headless-cms/components/cmsSidebarContent/CmsSidebarContentLayoutLinkList.tsx
```diff
@@ -0,0 +1,67 @@
...
+const CmsSidebarContentLayoutLinkList: React.FC<Props> = ({
+  anchor,
+  title,
+  description,
+  links,
+}) => {
+  return (
+    // I assumed that the anchor references the ID that an anchor element may
+    // target with href="#anchor"
+    <div id={anchor || undefined}>
```
Sartonon, line 34: What is this used for?

Sartonon (approved):
Looks good :)

src/headless-cms/components/cmsSidebarContent/CmsSidebarContentLayoutLinkList.tsx
Jokinen: Honestly, I'm not sure. My assumption is, because of the data model, that it's an ID that can be used create anchor links.

The data model seems to be more suited for creating sections. In which case linking to them makes more sense. E.g. you could have a link within content that links to a section later on ("Check out this content").

Here it doesn't make as much sense, but I added it to be complete in case it's not in this data model by coincidence, but because there's actual some requirement it checks.

Jokinen:
Resolved conflicts with base branch.

PR #222, opened 2022-01-26
PT-1450 | Allow content creators to highlight articles and pages in the sidebar

Jokinen:
## Description :sparkles:

Content creators can add article and page collections into the sidebar.

The articles and pages are extracted from the collections, with all collection meta information being ignored. This is in line with the implementation in nuorten.helsinki. This means title, description and all other such information is ignored.

If an article or page does not have an image, the title is wrapped in a coloured box instead.

### Closes :no_good_woman:

**[PT-1450](https://helsinkisolutionoffice.atlassian.net/browse/PT-1450)**

### Automated tests :gear:️

I've tested that pages and articles are rendered when provided in the data.

## Screenshots :camera_flash:

Screenshot 2022-01-26 at 8 07 30
Screenshot 2022-01-26 at 8 07 41


src/headless-cms/components/cmsSidebarContent/CmsSidebarContent.tsx
```diff
@@ -1,32 +1,95 @@
...
+
+function createCmsLink(cmsUri: string, language: Language) {
+  const baseUri = `/cms-page${cmsUri}`;
+
+  if (language === 'fi') {
+    return baseUri;
+  }
+
+  return `/${language}${baseUri}`;
+}
```
Sartonon, line 95: This implementation has problem when linked to pages with other language, for example the following url is generated when linking to english page:

`/en/cms-page/en/learning-materials/elementary-school`

There is also `getCmsPath` util in this project tat tries to accomplish the same task

```
const getCmsPath = (slug?: string | null): string => {
  return ROUTES.CMS_PAGE.replace('/:slug', slug ?? '');
};
``` 

Maybe `getCmsPath` could be refactored little bit and used also here. If `getCmsPath` would receive locale as second parameter some other places need refactoring too, but I think it is the better solution.

Sartonon (changes requested):
UI and code looks very good. Just need to fix bug with `createCmsLink` and I also created PR with typescript fix in test file.

src/headless-cms/components/cmsSidebarContent/CmsSidebarContent.tsx
Jokinen: I checked out your suggestion, but understood while doing it that this old utility was alone as is when combined with `stripLocaleFromUri` from `headless-cms` utils.

The url gets passed on to a `next/link` component that adds the correct locale into the link.

Sartonon (approved):
Great 👍

PR #224, opened 2022-01-28
PT-1476 | Handle situation when cms page is not found

Sartonon:
## Description :sparkles:

Handle fallback situation in app.tsx to avoid strange behaviour when cms page is not found

### Closes :no_good_woman:

**[PT-1476](https://helsinkisolutionoffice.atlassian.net/browse/PT-1476):**

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -3,8 +3,11 @@ import {
   MatomoProvider,
 } from '@datapunt/matomo-tracker-react';
 import * as Sentry from '@sentry/browser';
+import { LoadingSpinner } from 'hds-react';
 import { appWithTranslation } from 'next-i18next';
 import { AppProps } from 'next/app';
+import NextError from 'next/error';
+import { useRouter } from 'next/router';
 import React, { ErrorInfo } from 'react';
 import { Provider } from 'react-redux';
 import { ToastContainer } from 'react-toastify';
@@ -41,6 +44,7 @@ const matomoInstance = createMatomoInstance({
 });
 
 const MyApp = ({ Component, pageProps }: AppProps) => {
+  const router = useRouter();
   const locale = useLocale();
   const cmsApolloClient = useCmsApollo(pageProps.initialApolloState);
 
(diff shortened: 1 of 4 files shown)


src/pages/_app.tsx
```diff
@@ -89,5 +103,21 @@ class ErrorBoundary extends React.Component {
...
+const Center: React.FC = ({ children }) => {
+  return (
+    <div
+      style={{
+        width: '100vw',
+        height: '100vh',
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+      }}
```
Jokinen, line 115: Minor comment here. This was probably left in the Liikunta app in accident when the solution was being PoC. In that sense I don't recommend using a style prop here, but I'm also perfectly fine with it as it only wraps the loading spinner.

src/pages/cms-page/[...slug].tsx
```diff
@@ -54,13 +54,62 @@ export async function getStaticPaths() {
...
+    return {
+      props: {
+        error: {
+          statusCode: 400,
+        },
+      },
+      revalidate: 10,
+    };
+  }
+}
```
Comment (Jokinen) [inline src/pages/cms-page/[...slug].tsx:110]: I always wondered whether there was a way to declare this logic so that it could be composed and pages could share it. For example:

```js
function computeStaticProps(executeFunction, config) {
  try {
    executeFunction()
  } catch (e) {
    console.error()
    return {} // config
  }
}
```

I wasn't confident enough in that it made sense so I just inlined everything and committed a bit of copy pasting.

In this application it's really not a necessity considering you are only declaring this logic once. But I'm still wondering whether you had any similar thoughts?

Jokinen (approved):
Wrote down something that popped into my mind while looking through these changes.

Tested that the issue is fixed 👍

src/pages/_app.tsx
Sartonon: I understand your concern but I'm also fine with this style props in this case.

src/pages/cms-page/[...slug].tsx
Comment (Sartonon) [reply src/pages/cms-page/[...slug].tsx:110]: I didn't exactly have those thoughts because just like you said in this application there are no similar `getStaticProps` implementations. My main reason was to avoid too long try catch section and also to make it little bit easier to understand.

I think sharing logic might make sense in liikunta.

PR #235, opened 2022-04-11
feat(search): added a isFree -filter to the events search

nikomakela:
PT-1437 PT-1440

3 results without filter (2 of them are free):
image

When the isFree checkbox is checked, only 2 results are visible (and they both are free)
image

### DIFF
--- a/src/domain/events/eventSearchForm/EventSearchForm.tsx
+++ b/src/domain/events/eventSearchForm/EventSearchForm.tsx
@@ -4,6 +4,7 @@ import { Button, IconSearch } from 'hds-react';
 import { useTranslation } from 'next-i18next';
 import React from 'react';
 
+import CheckboxField from '../../../common/components/form/fields/CheckboxField';
 import DateInputField from '../../../common/components/form/fields/DateInputField';
 import DivisionSelectorField from '../../../common/components/form/fields/DivisionSelectorField';
 import MultiDropdownField from '../../../common/components/form/fields/MultiDropdownField';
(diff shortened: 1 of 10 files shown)


src/domain/events/eventSearchForm/EventSearchForm.tsx
```diff
@@ -170,57 +177,69 @@ const EventSearchForm = ({
...
+                      options={languageOptions}
+                    />
+                  </div>
+                  <div className={styles.filtersRow4}>
+                    <div className={styles.checkboxWrapper}>
+                      <Field
+                        component={CheckboxField}
+                        name="isFree"
+                        label={t('events:search.labelIsFree')}
+                        style={checkboxSearchInputStyles}
```
Jokinen, line 238: Other fields in this component seem to be styled with `className`. Should this field be styled in the same way?

src/domain/events/eventSearchForm/eventSearchForm.module.scss
```diff
@@ -70,6 +70,17 @@
   }
 }
 
+.checkboxWrapper {
+    align-items: center;
```
Jokinen, line 74: Seems like there's a 4 space indent here when the other rows in the file use 2.

Prettier should also support scss files.

Is this an issue with how the project has been configured considering that these haven't printed warnings?

src/domain/events/eventSearchForm/eventSearchForm.module.scss
```diff
@@ -70,6 +70,17 @@
...
+    & > div > input {
+      border: none !important;
+      background: var(--color-white);
+    }
+    & > div > label {
+      color: var(--color-white) !important;
+    }
```
Jokinen, line 81: If you apply a className to the checkbox component, you can overwrite these styles by using CSS variables:

```css
.className {
  --label-color: ...;
  --background-unselected: ...;
  ...
}
```

This approach has a bit clearer contract between HDS and this service.

Minor.

Might be too verbose for your taste.

Jokinen (approved):
Finding the price details for non-free events takes some clicking to do!

This change seems to be working nicely 👍 Minor code style comments.

PR #238, opened 2022-04-14
PT-1438 | Save EventSearchPage sort state to localStorage

charn:
## Description :sparkles:

Selected sort state for the list of events is persisted.

### Closes :no_good_woman:

**[PT-1438](https://helsinkisolutionoffice.atlassian.net/browse/PT-1438):** As a teacher, when browsing events, I want "last modified" order to stay if I have chosen it

**TODO**
- [ ] Add tests


Jokinen (approved):
Looks good 👍

Tested the behaviour and it works. Order selection persists over tab closure or navigation to a different page.

PR #249, opened 2022-10-27
PT-1560 | Add terms of service to the footer

tuomas777:
## Description :sparkles:

Added terms of service CMS page to the footer.

### Closes :no_good_woman:

**[PT-1560](https://helsinkisolutionoffice.atlassian.net/browse/PT-1560): Add the terms of service to the footer**

## Screenshots :camera_flash:

Screenshot 2022-10-27 at 23 44 35


src/constants.ts
```diff
@@ -56,6 +56,12 @@ export const PROVIDER_UI_LINKS = {
   sv: 'https://provider.kultus.fi/sv',
 };
 
+export const TERMS_OF_SERVICE_SLUGS = {
+  fi: 'kayttoehdot',
+  en: 'terms-of-service',
+  sv: 'anvandarvillkor',
+};
```
nikomakela, line 63: Would be better if it could also be configurable with env-variables. Just for your information, it's also possible to use the same slug for all the language variations, but I think it's better to use locale-specific URL's like you have used here 👍

nikomakela (approved):

nikomakela:
The PR-server goes to 404 https://palvelutarjotin-ui-pt-1560-add-terms-of-s-94.test.kuva.hel.ninja/cms-page/terms-of-service 
image

src/constants.ts
tuomas777: Added env support for those.

tuomas777:
> The PR-server goes to 404 https://palvelutarjotin-ui-pt-1560-add-terms-of-s-94.test.kuva.hel.ninja/cms-page/terms-of-service 

Yeah, apparently for some reason the review envs are using production headless CMS, which does not have the terms of service pages yet.

PR #251, opened 2022-11-03
feat: dynamic footer links from cms

nikomakela:
PT-1566.

The missing pages are added to the staging and production environments of the Headless CMS. All the footer link items are now fetched from the headless CMS.

I needed to update some dependencies also because the Kultus failed to jest-test the new feature (and it's imports).

NOTE: Footer could have a test, but I couldn't get the mock provider to work. The test passes with longer timeout, but it renders the actual content from the headless cms, so I skipped the test and added a todo-comment.

image


nikomakela:
Was a bit bigger than 2 points and 1 hook 😂

melniiv (approved):

PR #258, opened 2022-12-21
feat: cookie consent implementation

melniiv:
### Closes :no_good_woman:

### DIFF
--- a/src/domain/cookieConsent/CookieConsent.tsx
+++ b/src/domain/cookieConsent/CookieConsent.tsx
@@ -56,18 +56,32 @@ const CookieConsent: React.FC<Props> = ({ appName, allowLanguageSwitch }) => {
                 expiration: t('common:consent.expiration.year'),
               },
               {
-                id: 'matomo',
-                name: '_pk*',
-                hostName: 'digia.fi',
-                description: t('common:consent.cookies.matomo'),
-                expiration: t('common:consent.expiration.days', { days: 393 }),
-              },
-              {
                 id: 'i18next',
                 name: 'i18next',
                 hostName: 'api.hel.fi',
                 description: t('common:consent.cookies.i18next'),
                 expiration: t('common:consent.expiration.session'),
+              },
+            ],
+          },
+        ],
+      },
+      optionalCookies: {
+        groups: [
+          {
+            title: t('common:consent.groups.matomo.title'),
+            text: t('common:consent.groups.matomo.text'),
+            expandAriaLabel: t('common:consent.groups.matomo.expandAriaLabel'),
+            checkboxAriaDescription: t(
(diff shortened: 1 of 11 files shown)


src/domain/cookieConsent/CookieConsent.tsx
```diff
@@ -0,0 +1,127 @@
...
+                id: 'servicemap_ga',
+                name: '_ga*',
+                hostName: 'palvelukartta.hel.fi',
+                description: t('common:consent.cookies.servicemap_ga'),
+                expiration: t('common:consent.expiration.session'),
+              },
+            ],
+          },
+        ],
+      },
```
nikomakela, line 112: Only thing I found related to Palvelukartta, is a 1 link that we create, but which is also handled as an external link that opens the target in a new tab. I don't think we need to have an optional cookie option for that, do we? Because we are already indicating that it's opening in a different site. If it is clear enough, that I don't know. Anyway, we don't even have feature that would be triggered from the selection. Should the link be hidden if the cookie consent is not accepted?

nikomakela (commented):
Looks good! I tested that 
- the cookie consent is not shown when it's accepted in another City of Helsinki site like Events ✅ 
- the language change effects on the cookie consent too ✅ 
- the cookie consent appears until user has reacted to it ✅ 

I would just remove (or comment out or add a feature flag or something to) the Palvelutarjotin optional cookie section, because we are not using the Palvelukartta cookie anywhere - we are just creating an external link to the Palvelukartta. However, it might be that we need the Palvelukartta in near future.
image

src/domain/cookieConsent/CookieConsent.tsx
nikomakela: The places menu is actually using the Palvelukartta through the Unified-Search, but it's still not using it's cookies.

melniiv: yes, fixed

nikomakela (approved):

melniiv:
moved matomo to optional cookies (based on new spex), updated translations, fixed tests.

src/domain/matomo/Matomo.tsx
```diff
@@ -34,15 +35,29 @@ function Matomo({ children }: { children: React.ReactNode }): JSX.Element {
...
   // Track page changes when pathnname changes
   useEffect(() => {
+    const getConsentStatus = (cookieId: string) => {
+      const consents = getAllConsents();
+      return consents[cookieId];
+    };
+
+    //if enabled, should be callled before each trackPage instruction
+    if (getConsentStatus('matomo')) {
+      console.log(123);
```
nikomakela, line 51: console statement

nikomakela (approved):
Approvable when the console statement is removed

PR #287, opened 2023-09-01
PT-1686 | feat: move "Enrol to queue" button under the occurrences table

karisal-anders:
### feat: move "Enrol to queue" button under the occurrences table

refs PT-1686

### Closes :no_good_woman:

**[PT-1686](https://helsinkisolutionoffice.atlassian.net/browse/PT-1686)**

## Testing :alembic:

https://kultus-ui-pr287.dev.hel.ninja/

## Additional notes :spiral_notepad:

[PT-1686]: https://helsinkisolutionoffice.atlassian.net/browse/PT-1686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #288, opened 2023-09-05
Move enrol to queue button below the occurrences table and hide it when queue should not be possible

nikomakela:
PT-1686
Move enrol to queue button below the occurrences table. Change the style and the text of the heading.

PT-1678
The event queue form should be hidden when there are no occurrences

PT-1677
The event queue form should be available only for the events with internal enrolment when the enrolment period has already started.

Second try for #287 .

----

### The event queue section is below the occurrences table when a single internal enrolment is required

image

The form can be opened and used

image

### The event queue section is below the occurrences table when multiple internal enrolments are required

image

### The queue enrolment controls are hidden when the event has external enrolments:

image

### The event queue section is hidden when the event has no enrolments

image

### The event queue section is hidden when the event has no occurrences

image

### DIFF
--- a/src/domain/event/__tests__/EventPage.test.tsx
+++ b/src/domain/event/__tests__/EventPage.test.tsx
@@ -976,3 +976,124 @@ describe('back button', () => {
     expect(backLink).toHaveAttribute('href', '/');
   });
 });
+
+describe('event queue enrolment form', () => {
+  it('is visible when the internal enrolments has already begun', async () => {
+    renderComponent();
+    await waitForRequestsToComplete();
+    expect(
+      screen.getByRole('heading', {
+        name: /jonoon ilmoittautuminen/i,
+      })
+    ).toBeInTheDocument();
+    expect(
+      screen.getByRole('button', {
+        name: /ilmoittaudu jonoon/i,
+      })
+    ).toBeInTheDocument();
+  });
+
+  it('is hidden when the internal enrolments has not yet begun', async () => {
+    const today = new Date();
+    const tomorrow = new Date(today);
+    tomorrow.setDate(tomorrow.getDate() + 1);
+    const eventStillNotEnrollable =
+      createEventQueryMockIncludeLanguageAndAudience({
+        ...eventData,
+        id: data.id,
+        pEvent: fakePEvent({
+          ...eventData.pEvent,
+          enrolmentStart: tomorrow,
+        }),
+      });
+    const [, ...mocks] = apolloMocks;
+    renderComponent({ mocks: [eventStillNotEnrollable, ...mocks] });
+    await waitForRequestsToComplete();
+    expect(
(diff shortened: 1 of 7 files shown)


terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr288.dev.hel.ninja** :rocket::rocket::rocket:

nikomakela:
The product owner checked this PR and all it's redesigns that were shown in the screenshots -- and they were OK! 🚢

src/domain/event/utils.ts
```diff
@@ -169,3 +170,20 @@ export const getEnrolmentType = (
   }
   return EnrolmentType.Unenrollable;
 };
+
+/**
+ * Does the event support enrolling in the queue?
+ * 1. Are the enrolments handled internally in the Kultus
+ * 2. Has the enrolment started already
+ * 3. Are there any occurrences for the event
+ */
```
karisal-anders, line 179: How would you feel about this:
```typescript
/**
 * Does the event support enrolling in the queue?
 * @return True if enrolments are handled internally in Kultus,
 * the enrolment has started and there is at least one occurrence
 * for the event, otherwise False.
 */
```

src/domain/event/__tests__/EventPage.test.tsx
```diff
@@ -976,3 +976,124 @@ describe('back button', () => {
...
+    const eventWithoutOccurrences =
+      createEventQueryMockIncludeLanguageAndAudience({
+        ...eventData,
+        id: data.id,
+        pEvent: fakePEvent({
+          ...eventData.pEvent,
+          occurrences: undefined,
+        }),
+      });
+    const [, ...mocks] = apolloMocks;
```
karisal-anders, line 1085: Unneeded comma?

karisal-anders (approved):
Approved with comments, please would you go through them? Thank you!

src/domain/event/utils.ts
nikomakela: Quite trivial thing. I think that a list is more clear, but I can still change this, just for the `@return ` jsdoc syntax 😄

terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr288.dev.hel.ninja** :rocket::rocket::rocket:

src/domain/event/__tests__/EventPage.test.tsx
karisal-anders: Ah, learned something new then. 👍 Looked up https://javascript.info/destructuring-assignment

src/domain/event/utils.ts
karisal-anders: To myself the list was otherwise clear but whether the bullet points meant OR or AND was not clear and that's why I suggested and alternative.

PR #306, opened 2024-02-21
feat: use keywordAnd-filter in events list query for keywords

nikomakela:
PT-1591.
Update the GraphQL schemas: Add KeywordAnd-filter to events list query.
The new keywordAnd filter is now used instead of the keyword-filter,
which means that all the selected keywords narrows down 
the search results list instead of growing it.

There is an alternative or "extension" for this in https://github.com/City-of-Helsinki/palvelutarjotin-ui/pull/307

### DIFF
--- a/src/common/components/keyword/KeywordsList.tsx
+++ b/src/common/components/keyword/KeywordsList.tsx
@@ -32,8 +32,8 @@ const KeywordsList: React.FC<KeywordsListProps> = ({
 
   return (
     <ul className={styles.keywordsList}>
-      {keywordsPropArr.map((k) => (
-        <li key={k.id}>
+      {keywordsPropArr.map((k, i) => (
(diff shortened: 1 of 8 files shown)


terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr306.dev.hel.ninja** :rocket::rocket::rocket:

karisal-anders:
Do you see anything that is better in this PR than in PR #307? If not, how about just closing this as unneeded/unused/replaced?

nikomakela:
> Do you see anything that is better in this PR than in PR #307? If not, how about just closing this as unneeded/unused/replaced?

We will continue with keyword_AND -parameter while we are waiting for LE-team to inestigate the issues related to in use of keyword_OR_set -parameter.

PR #327, opened 2024-09-20
fix: use only the values of a persisted form state when initializing …

nikomakela:
…a form with FormikPersist

PT-1721.

The FormikPersist persists the whole Formik state to the local storage, so it included all the errors and dirtiness checks. When user opened the form again, if there were any errors in the persisted form state, the form also had all the validation errors visible. The validation errors replaces the helper texts. Instead of initializing the full form state with the persisted state, use only the persisted form values.

image


karisal-anders (approved):
Looks ok codewise, didn't test, let's test in test environment, and fix if needs fixing.

terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr327.dev.hel.ninja** :rocket::rocket::rocket:

PR #328, opened 2024-09-23
fix: group size field validation when no limits are set

nikomakela:
PT-1772.

The group sizes can be left empty, especially when 1 group is set to fill the event. In those cases, the group size children and adult inputs should not validate themselves against the unset limits. The fields should only be required when the limits are not set.
image

### DIFF
--- a/src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
+++ b/src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
@@ -3,8 +3,8 @@ import * as Yup from 'yup';
 import { VALIDATION_MESSAGE_KEYS } from '../../app/forms/constants';
 
 type EnrolmentFormValidationSchemaProps = {
-  minGroupSize: number;
-  maxGroupSize: number;
+  minGroupSize?: number;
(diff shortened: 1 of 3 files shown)


src/domain/enrolment/enrolmentForm/ValidationSchema.tsx
```diff
@@ -55,10 +55,12 @@ export default function getValidationSchema({
           // The used validation error message will be the same for both the fields.
           // This is also preventing negative param.max to occur in validation.
           if (
-            !sizePair ||
-            sizePair < 0 ||
-            sizePair < minGroupSize ||
-            sizePair > maxGroupSize
+            minGroupSize &&
+            maxGroupSize &&
```
karisal-anders, line 59: Is it possible to define only `minGroupSize` or `maxGroupSize` but leave the other undefined? If that is possible, then should that case also be taken into account here?

karisal-anders (approved):
Please take a look at the comment I left, and whether we should handle the possible cases where only `minGroupSize` or `maxGroupSize` is defined, but not both. Looks ok to merge besides that, let's test more in test environment.

PR #332, opened 2024-09-25
Enrolment status in event card

nikomakela:
Dependency to https://github.com/City-of-Helsinki/palvelutarjotin/pull/365 which gives a new field to help to resolve the status code for the card.

PT-1668.

The enrolment status of an event is shown in the event card. The event
card can now have 4 different statuses:
- The event has some space available in some of it's occurrences.
- The event is full, it does not have any space left in any of it's
occurrences.
- The event enrolment status could not be resolved, which is when the
event has external enrolments.
- The event has no enrolments enabled at all in any of it's occurrences.

New important enrolment status fields are added to events fields query fields fragment.

NOTE: Also updated the CMS schema as it was out of the date a bit.

Screenshot 2024-09-24 at 11 54 08


terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr332.dev.hel.ninja** :rocket::rocket::rocket:

public/locales/en/event.json
```diff
@@ -76,6 +76,11 @@
     "deleteDateFilter": "Poista rajaus"
   },
   "keywords": {
-    "labelKeywordLink": "Search events with keyword {{keyword}}"
+    "labelKeywordLink": "Search events with keyword {{keyword}}",
+    "enrollmentStatus": {
+      "externally_unknown": "Registrering på en annan sida",
+      "free_space_available": "Platser kvar",
+      "full": "Full"
+    }
```
karisal-anders, line 84: enrollmentStatus in en & sv, enrolmentStatus in fi,
enrolmentStatus would be in line with the British English use in the backend i.e. enrolment (British English) instead of enrollment (American English). Either way a single name should be used in en/sv/fi json files.

melniiv (approved):

src/common/components/keyword/keyword.module.scss
```diff
@@ -10,6 +10,24 @@
...
 }
+.keyword-red {
+  background-color: var(--color-brick);
+  color: white;
+}
+
+.keyword-green {
+  background-color: var(--color-copper-dark);
+  color: white;
+}
```
karisal-anders, line 21: Are the keyword colors actually categories that have nothing to do with color? I mean are they actually semantic categories, not color categories? If they are color categories then should the color of the category's name be reflected in the actual color shown? We could have a chat about this too, maybe easier to talk live / throw ideas around.

karisal-anders: If they're semantic categories then I'd say renaming the categories to their semantic meanings would be better than having the semantic information mapped to color in the name and still mapped to some other color in the visible shown color, if I understood correctly what's going on here. But let's talk!

nikomakela: I thought about this and I personally came to conclusion that the color-as-a-identifier told me more than semantic. I can think that green is a good thing and there is something good for me, but if the name is something like "free", it does not tell me the color of it and I might not even understand the whole "labeling". But, if this is something that really matters to you and you come up with better naming, it's OK for me to change the implementation. I think this is very trivial thing after all.

nikomakela: Also, this styling is now in keyword context, not in enrolment status context.

karisal-anders: We had a talk and this is ok.

terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr332.dev.hel.ninja** :rocket::rocket::rocket:

PR #336, opened 2024-10-08
fix: set CMS page search size to better fit on grid implementation

nikomakela:
PT-1771.

Instead of fetching 10 items to a grid that has 3 columns, fetch 9 items, so they are equally balanced in the grid.

image


terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr336.dev.hel.ninja** :rocket::rocket::rocket:

PR #345, opened 2024-11-03
fix: sentry to all envs enabled

melniiv:
### Closes :no_good_woman:


terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr345.dev.hel.ninja** :rocket::rocket::rocket:

nikomakela (changes requested):
Why is this needed? It is the `NODE_ENV` (not the pipeline env), which can be either `"development"` or `"production"`. The `NODE_ENV` should be `"development"` only in localhost when you are running a development server, but not even there, if you are using a docker container with a production image `DOCKER_TARGET=production docker compose up --build`. I think it is a bad thing to initialize the Sentry to be in use in my local machine!

PR #349, opened 2024-12-04
Event page mobile payout

melniiv:
## Description :sparkles:
The layout was broken due to the wrong markup in multiple components
1. Hero: mobile markup should be completely different with opposite positioning for child containers
2. Occurrences section: table and time fields positioning refactored. Table now has a horizontal scroll, that was the main component which was breaking layout and header specifically
3. Queue: repositioned in mobile
4. Share:  repositioned in mobile
5. Removed huge margin bottom in mobile between content and footer.

Please test the desktop and mobile view.
Still question, how usable the table is on mobile. User has to scroll to see the Details button, and then when details are expanded, user sees the most right content.

Hero in mobile:

image

Header expanded:
image

Mobile content of the event page:

image

With scrolled table to right:
image

With expanded details: 
image

Tos see other expanded details, should be scrolled back to the left.

### Closes :no_good_woman:

**[PT-1829](https://helsinkisolutionoffice.atlassian.net/browse/PT-1829):**

[PT-1829]: https://helsinkisolutionoffice.atlassian.net/browse/PT-1829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/common/components/table/table.module.scss
+++ b/src/common/components/table/table.module.scss
@@ -2,6 +2,8 @@
 
 .tableWrapper {
   max-width: 100%;
+  overflow-x: auto;
+  white-space: nowrap;
 }
 
 .table {
@@ -15,6 +17,7 @@
   --fontsize-cell-header: var(--fontsize-body-l);
   --margin-bottom-table: var(--spacing-m);
   --spacing-cell: var(--spacing-m) var(--spacing-xs);
+  --spacing-cell-mobile: var(--spacing-xs) var(--spacing-xs);
   --expanded-row-border-color: var(--color-black-30);
 
   width: 100%;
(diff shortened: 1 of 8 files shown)


terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr349.dev.hel.ninja** :rocket::rocket::rocket:

karisal-anders (commented):
Around ~960px the mobile menu is still available:
- image

But around ~930px it's partially off screen:
- image

And at ~900px the menu is not available on screen:
- image

The menu continues to be off screen until ~575px:
- image
- image

So the **menu** is **not available** on screen between ~**900px–575px** width.

nikomakela (approved):
Like you said @melniiv, the table is not very good on mobile, but everything else except that works quite well on my Google Pixel 8 and also in my Pixel tablet.

@karisal-anders Some of the widths might be calculated on page render and those are not responsive when the display size is changed without reloading. I think that it is acceptable, since it is clear that the mobile view has not been any of the priorities. I personally think that the mobile view is important, but I do remember conversations with designers where they have stated that it is not and the UI is aimed for big laptop screens.

However, I don't think the "Ilmoittaudu jonoon" should be centered:
image

melniiv:
fixed the "Ilmoittaudu jonoon"

melniiv:
the problem that Kari mentioned is related to the footer links, i ll create a separate ticket for that

PR #359, opened 2025-03-07
fix: rehydration issue in event cards keywords and enrolment statuses

nikomakela:
PT-1844.

The actual rehydration issue came from the useTranslation hook, which was not imported from the next-i18next, but from the react-i18next. Because of that, the result some how differed between the renders (from SSR and client).

The EnrolmentStatusKeyword was also rendered without LI-element, even then when it was rendered inside an UL-element. That is invalid against HTML standard.

Also, some key-attributes are now added with some uniqueID prefixes, because otherwise React struggled to identify the unique keyword items. This could have been an issue during rerenders. It was warned in the console.

### DIFF
--- a/src/common/components/keyword/EnrolmentStatusKeyword.tsx
+++ b/src/common/components/keyword/EnrolmentStatusKeyword.tsx
@@ -1,4 +1,4 @@
-import { useTranslation } from 'react-i18next';
+import { useTranslation } from 'next-i18next';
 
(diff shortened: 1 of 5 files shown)


terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr359.dev.hel.ninja** :rocket::rocket::rocket:

karisal-anders (approved):

PR #374, opened 2025-03-21
Fix notification query client usage

nikomakela:
PT-1795.

There was an issue in CMS notification query, because it was using a
wrong Apollo Client. It was using the one meant not for Headless CMS but
for Kultus API.

### DIFF
--- a/src/domain/app/layout/__tests__/PageLayout.test.tsx
+++ b/src/domain/app/layout/__tests__/PageLayout.test.tsx
@@ -1,7 +1,6 @@
-import { MockedResponse } from '@apollo/client/testing';
 import React from 'react';
 
-import { NotificationDocument } from '../../../../generated/graphql-cms';
+import * as GraphQLCms from '../../../../generated/graphql-cms';
 import { emptyMenuQueryMocks } from '../../../../tests/apollo-mocks/menuMocks';
 import { fakeNotification } from '../../../../utils/cmsMockDataUtils';
 import {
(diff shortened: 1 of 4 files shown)


src/domain/app/layout/__tests__/PageLayout.test.tsx
```diff
@@ -13,54 +12,53 @@ import {
...
+    loading: false,
+  } as any);
   const { container } = render(
     <PageLayout>
       <div>Page layout children</div>
     </PageLayout>,
-    { mocks: mocksWithNonEmptyNotification }
+    { mocks: [...emptyMenuQueryMocks] }
   );
   expect(container.firstChild).toMatchSnapshot();
```
karisal-anders, line 43: Why did the snapshot here not change as before there was a non-empty notification here and now there is an empty notification here? PageLayout → Header → HeaderNotification → Notification so it should be there, or is e.g. PageLayout's render's firstChild somehow omitting it / stepping into some part of the layout that does not contain the notification?

terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr374.dev.hel.ninja** :rocket::rocket::rocket:

src/domain/app/layout/__tests__/PageLayout.test.tsx
karisal-anders: PageLayout.test.tsx.snap doesn't have any notification data in it so that's why it's matching. Maybe the non-empty notification mock didn't really work before?

karisal-anders (approved):
Please take a look at the comment https://github.com/City-of-Helsinki/palvelutarjotin-ui/pull/374#discussion_r2007501492 I left, otherwise looks ok

src/domain/app/layout/__tests__/PageLayout.test.tsx
nikomakela: the mocks included menu and notification queries. Now the mocks includes only the menu query but the notification is mocked with spy. The result is exactly the same. The spy was needed, because now when I have given the client as an attribute to the query, it's not the same as the mocked apollo client in tests.

karisal-anders: We had a talk, and we came to a consensus that there was probably something wrong here. But it is ok to have the empty notification in this test case as there's a test case for the notification being there and not being there later.

PR #381, opened 2025-03-27
feat(csp): add CSP header configuration

nikomakela:
PT-1874.

See more:
https://nextjs.org/docs/14/pages/building-your-application/configuring/content-security-policy

PR #382, opened 2025-03-27
feat(csp): middleware implementation for CSP

nikomakela:
PT-1874 PT-1893.

To support usage of nonces, a middleware must be used. Reference:
https://nextjs.org/docs/app/building-your-application/configuring/content-security-policy.

However, it seems like the nonces are not usable when using ISR, because the static page generator cannot have a request specific nonce.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
 import type { NextRequest } from 'next/server';
 import { NextResponse } from 'next/server';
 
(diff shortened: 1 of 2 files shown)


nikomakela:
This PR is created more like as an example. I self prefer the implementation done in #381, since it's so much simpler and easier to maintain.

terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr382.dev.hel.ninja** :rocket::rocket::rocket:

PR #384, opened 2025-04-02
Fix event details back button scroll restoration

nikomakela:
PT-1886.

Steps to test:
1. Search an event
2. Click an event card to navigate to event details page
3. Use browser back button or back arrow -button in UI to navigate back to search page
Result: The serach results should be auto loaded (with Apollo cache) and user should be taken to the clicked card.

### DIFF
--- a/src/domain/event/BackFromEventDetailsButton.tsx
+++ b/src/domain/event/BackFromEventDetailsButton.tsx
@@ -5,18 +5,30 @@ import { useTranslation } from 'next-i18next';
 import React from 'react';
 
 import { createBackButtonHref } from './utils';
+import { SCROLL_RESTORATION_ELEMENT_ID_PREFIX } from '../events/constants';
 
 /**
  * Back to event search from the event details page.
  */
 export default function BackFromEventDetailsButton() {
   const { t } = useTranslation();
-  const { asPath } = useRouter();
-  const [, search] = asPath.split('?');
-  const href = createBackButtonHref(search);
+  const {
+    asPath,
+    query: { eventId },
+  } = useRouter();
+  const [, currentSearch] = asPath.split('?');
+  const { pathname, search } = createBackButtonHref(currentSearch);
+  const hash = `#${SCROLL_RESTORATION_ELEMENT_ID_PREFIX}${eventId}`;
 
+  const href = pathname + (search ? `?${search}` : '') + hash;
+  // Disable back button scroll, because we have our own custom implementation
+  // to fix issues in nextjs: see
+  // 1. https://github.com/vercel/next.js/issues/68746
+  // 2. https://github.com/vercel/next.js/issues/3303
+  // 3. https://github.com/vercel/next.js/discussions/44013
+  const scroll = false;
   return (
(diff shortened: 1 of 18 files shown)


karisal-anders:
Some of the static tests are timing out, I noticed that while doing the major upgrading recently with palvelutarjotin-ui that sometimes some of the static tests did that. Just adding some more leeway to the specific timing out tests would probably work as an immediate band-aid at least.

src/domain/events/EventsSearchPage.tsx
```diff
@@ -77,6 +77,31 @@ const EventsSearchPage: React.FC = () => {
...
+        )
+        .catch((e) => {
+          // workaround for https://github.com/vercel/next.js/issues/37362
+          if (!e.cancelled) {
+            throw e;
+          }
+        });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
nikomakela, line 103: Could be moved to a new hook.

karisal-anders (approved):
Tested locally but only with 3 events so didn't test the load more case. Worked ok with both browser back button (then it went to the exact place where it left) and the back button shown on the event details page (then it went to the event card that was related to the opened event details page).

src/domain/event/BackFromEventDetailsButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Back to event search from the event details page.
+ */
+export default function BackFromEventDetailsButton() {
+  const { t } = useTranslation();
+  const {
+    asPath,
+    query: { eventId },
+  } = useRouter();
+  const [, currentSearch] = asPath.split('?');
```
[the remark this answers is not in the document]
nikomakela: I also tried useSearchParams, but the thing is that it is a client only component and it's only for NextJS App-router. It does not work properly in pages-router. The whole next/navigation package is for app-router. I'm sorry I cannot find a real source for this, but I tried to use it and it didn't work as I expected it to work. Here are at least some claims about this (eventhough not official ones):
- https://stackoverflow.com/a/77651688/784642
- https://github.com/vercel/next.js/issues/54242
One thing is that it is not even mentioned in [pages-router documentation](https://nextjs.org/docs/pages/api-reference/functions) , but is mentioned in [app-router docs ](https://nextjs.org/docs/app/api-reference/functions).

src/domain/events/EventsSearchPage.tsx
nikomakela: I added a new hook called "useHashlessReturnPathScrollFix"

nikomakela:
I found out that even though there is a unit test that tests that the search parameters stays when a back arrow navigation is executed, the search params are now getting cleared. This feature seems to still work in production env, so the issues might have come with these changes.

terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr384.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr384.dev.hel.ninja** :rocket::rocket::rocket:

PR #387, opened 2025-04-09
Refactor event card to be more accessible

nikomakela:
PT-1899.

Changed keyword list keywords to be static labels instead of link buttons.

Unwrap the event card link and use a pseudo-content trick instead. The primary link, meaning the event details link, is now set to a title and removed as a event card wrapper. This makes the content much more accessible, because the content can be picked to be not covered by a exposed link with `position: relative` CSS-property.

If there are any secondary links or buttons inside an event card, they are no longer wrapped by another links, making them more accessible too.

When focused or hovered, the card is now outlined with a small grey shadow, making it to stand up from others.

More about the solution and the issue for example in:
1. https://inclusive-components.design/cards/.
2. https://www.damianwajer.com/blog/accessible-card-component/
3. https://kittygiraudel.com/2022/04/02/accessible-cards/
4. https://medium.com/@chiaracielo/accessible-cards-by-design-4741e2158bb4
5. https://m3.material.io/components/cards/guidelines
6. https://hds.hel.fi/components/linkbox/
image


terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr387.dev.hel.ninja** :rocket::rocket::rocket:

karisal-anders (approved):
The commented out code in the tests could be removed unless there's a real reason to keep them, otherwise looks ok to me.

terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr387.dev.hel.ninja** :rocket::rocket::rocket:

PR #392, opened 2025-04-15
Fix search params handling on language change and refactor header component

nikomakela:
PT-1894.

The search filters were activated in search form properly, because the parameter names are case sensitive and the search params of the URL were lower cased.

Split the Header component code content to new hooks and util functions, so it's easier to manage and to understand, but also easier to test.

image

### DIFF
--- a/src/domain/headless-cms/__tests__/utils.test.ts
+++ b/src/domain/headless-cms/__tests__/utils.test.ts
@@ -1,5 +1,8 @@
+import { ParsedUrlQuery } from 'querystring';
+
 import isEqual from 'lodash/isEqual';
 import { graphql, HttpResponse } from 'msw';
+import { LanguageCodeEnum, MenuItem } from 'react-helsinki-headless-cms';
 import {
   MenuQuery,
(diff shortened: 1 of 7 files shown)


karisal-anders (approved):
Tested and retained parameters.

src/domain/headless-cms/__tests__/utils.test.ts
```diff
@@ -390,3 +414,298 @@ describe('rewriteInternalURLs', () => {
...
+      expect(getIsItemActive(item, locale)).toBe(expected);
+    }
+  );
+
+  it('should handle window being undefined (SSR)', () => {
+    const originalWindow = global.window;
+    // @ts-expect-error: Simulate SSR environment where window is undefined
+    delete global.window;
+    expect(getIsItemActive(menuItem, 'fi')).toBe(false);
+    global.window = originalWindow; // Restore window
```
karisal-anders, line 541: Might the temporary deletion of global.window interact badly with other tests being run if they were run just in the right moment? Would mocking it be safer here?

PR #395, opened 2025-04-17
Replace hard coded privacy statement links with instructions to find it from footer

nikomakela:
PT-1915.

image

image


karisal-anders (approved):
The different language versions vary a little now between these texts, the footer links and the documents accessible through the footer links. To make things clearer they could be made the same i.e. only one translation used to mean one thing in one language and not many. But it's ok as it is, but doing that would make things more transparent.

terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr395.dev.hel.ninja** :rocket::rocket::rocket:

PR #408, opened 2025-05-09
fix: only 1 enrolment notification should be shown at once

nikomakela:
PT-1800.

The notifications of enrolment queue and enrolment reservation query are controlled with URL search parameters. Sync the queue enrolment and real enrolment URL search parameters so that the 2 notifications are not shown at the same time.


terovirtanen:
**kultus-ui branch is deployed to platta: https://kultus-ui-pr408.dev.hel.ninja** :rocket::rocket::rocket:

karisal-anders (approved):

PR #425, opened 2025-09-29
fix: root index should apply configured headers

nikomakela:
The headers were set in nextjs configuration in a way that they were not applied in the root index page at all. All the rest of the pages did have the configured headers, but the root index did not. Now when the source has a different matcher, also the root index gets the configured headers.

Reference: HH-432.

image

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -172,8 +172,12 @@ module.exports = {
   async headers() {
     return [
       {
-        source: '/(.*)',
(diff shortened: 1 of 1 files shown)


nikomakela:
Reference to events-helsinki-monorepo PR: https://github.com/City-of-Helsinki/events-helsinki-monorepo/pull/844

Riippi (approved):

PR #489, opened 2026-05-25
fix: update headless cms HCRC-194

Riippi:
Refs: HCRC-194

## Description :sparkles:

Update react-helsinki-headless-cms from 3.0.0 to 3.0.3.
That contains dependency updates and bug fixes.

Also needed to add jest-environment-jsdom dependency. Before it was installed via something else, but now tests failed without it being installed.

Fixed two acceddibility issues:

<ul> had aria-labelled by pointing to a button that might not be in dom. Getting label from a button made no sense even if it was in dom.

Icon that was supposed to be hidden from screenreaders was not hidden because a prop was not properly passed.

**[HCRC-194](https://helsinkisolutionoffice.atlassian.net/browse/HCRC-194):**

[HCRC-194]: https://helsinkisolutionoffice.atlassian.net/browse/HCRC-194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accessibility by sanitizing generated checkbox/toggle DOM identifiers.
  * Removed an `aria-labelledby` reference that could point to an unsanitized toggle id.
* **New Features**
  * Added a DOM-id sanitization utility used across UI components.
* **Tests**
  * Added unit tests for sanitization behavior (safe vs. unsafe characters).
* **Refactor**
  * Updated the clock icon to use standard SVG props.
* **Chores**
  * Updated a CMS dependency and added the JSDOM Jest environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,8 +30,8 @@
     "fix-hds-shim": "node scripts/fix-hds-core-cookie-consent.js"
   },
   "dependencies": {
-    "@apollo/client": "^3.13.5",
-    "@city-of-helsinki/react-helsinki-headless-cms": "3.0.0",
+    "@apollo/client": "^3.14.0",
+    "@city-of-helsinki/react-helsinki-headless-cms": "3.0.3",
     "@date-fns/tz": "^1.2.0",
     "@jonkoops/matomo-tracker-react": "^0.7.0",
     "@sentry/nextjs": "^10.48.0",
@@ -104,6 +104,7 @@
     "jest": "^29.7.0",
     "jest-axe": "^10.0.0",
     "jest-date-mock": "^1.0.10",
+    "jest-environment-jsdom": "29.7.0",
     "jest-fixed-jsdom": "^0.0.9",
     "jest-localstorage-mock": "^2.4.26",
     "lint-staged": "^16.4.0",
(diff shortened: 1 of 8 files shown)


src/common/components/multiSelectDropdown/MultiSelectDropdown.tsx
```diff
@@ -336,7 +336,7 @@ const MultiSelectDropdown: React.FC<MultiselectDropdownProps> = ({
         <Checkbox
           ref={setFocus}
           checked={isChecked}
-          id={`${checkboxName}_${option.value}`}
+          id={`${checkboxName}_${option.value.replace(/[^a-zA-Z0-9_-]/g, '-')}`}
```
jorilindell: We could move the ID sanitizing logic into a shared helper that also accepts a prefix. Then we can call it like cleanId(option.value, checkboxName) and avoid repeating the regex in components.

```
export function cleanId(value: string, prefix?: string): string {
  const cleaned = value
    .replace(/[^a-zA-Z0-9_-]/g, '-')
    .replace(/-+/g, '-');

  return prefix ? `${prefix}_${cleaned}` : cleaned;
}
```

Riippi: Added similar id helper

package.json
```diff
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.13.5",
-    "@city-of-helsinki/react-helsinki-headless-cms": "3.0.0",
+    "@city-of-helsinki/react-helsinki-headless-cms": "3.0.3",
```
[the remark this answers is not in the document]
Riippi: Seems to work ok. React19 version coming soon.

tjgofore (approved):

mikkojamG (approved):

PR #495, opened 2026-06-02
PT-453 | Add cultural route support to enrolment form

karisal-anders:
## Description :sparkles:

Needs backend support from City-of-Helsinki/palvelutarjotin#509

**NOTE**:
- **SonarCloud's** **coverage** analysis **should** **be** **ok**, IsPartOfCulturalRouteField has tests for it, and the playwright page changes should not be counted against coverage. So, the coverage should be >65% even though it says not so.

### Add cultural route support to enrolment form

Also:
- with the request of the product owner don't restore any previous
  answer to "Is part of cultural route" question from local storage
  persisted Formik form's data, but instead always reset the question
  to its initial value i.e. no data given at all
- update generated GraphQL types using backend running
  City-of-Helsinki/palvelutarjotin#509 i.e. the backend support for the
  `is_part_of_cultural_route` field in enrolments and queue enrolments
- add tests
- update playwright tests

### Related :handshake:

- [PT-453](https://helsinkisolutionoffice.atlassian.net/browse/PT-453)
- Related **backend support**:
  - City-of-Helsinki/palvelutarjotin#509

## Screenshots :camera_flash:

See the screenshots from https://github.com/City-of-Helsinki/palvelutarjotin/pull/509 as this PR was used with the backend support in that PR.

[PT-453]: https://helsinkisolutionoffice.atlassian.net/browse/PT-453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/domain/enrolment/enrolmentForm/constants.ts
```diff
@@ -1,10 +1,16 @@
+export const enum IsPartOfCulturalRoute {
+  NO_OR_UNKNOWN = 'NO_OR_UNKNOWN',
+  YES = 'YES',
+}
```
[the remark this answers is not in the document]
karisal-anders: Hmm... at least when running `yarn build` + `yarn start` and trying the form validation it worked ok i.e. said "Tämä kenttä on pakollinen" when having not filled in the cultural route choice, and not saying that when it was chosen.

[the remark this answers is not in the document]
karisal-anders: Changed from `const enum` to `enum` with [force push](https://github.com/City-of-Helsinki/palvelutarjotin-ui/compare/f08c5ab6ea36f349928aece03afc1ba004bc1088..f37539f872f39bd3f28ad57464f36715069d2194) for clarity although it looked like it worked even without this change.

src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
```diff
@@ -206,6 +208,10 @@ const EnrolmentForm: React.FC<Props> = ({
               </FormGroup>
             </div>
 
+            <FormGroup>
+              <IsPartOfCulturalRouteField formikFieldName="isPartOfCulturalRoute" />
+            </FormGroup>
```
nikomakela, line 213: I don't think it changes the result or analysis any how, but now I could understand this is "No / I don't know what cultural route is / I don't want to give an answer". 

I would have personally used just a checkbox. If I know what a cultural route is, I would check it. I'm enrolling and I need to check that all the information that I have given with form is valid. If we would really trust in users and in the given input, this would be easy boolean analysis and way easier to implement and even cleaner UI and UX. Now we are doing much work only to guide the report analysis, because the input would still be the same -- This implementation still does not answer to report analysis question whether or not the user has seen this question.

mikkojamG (approved):
LGTM

src/domain/enrolment/enrolmentForm/__tests__/__snapshots__/EnrolmentForm.test.tsx.snap
```diff
@@ -112,6 +115,7 @@ exports[`renders form and user can fill it and submit and form is saved to local
     "hasEmailNotification": true,
     "hasSmsNotification": true,
     "isMandatoryAdditionalInformationRequired": false,
+    "isPartOfCulturalRoute": "YES",
```
nikomakela, line 118: so can it now be "",  " ", true, false, "YES" and "NO_OR_UNKNOWN"? But it's still converted to true / false when sent to graphql endpoint?

Then in backend, at least in reports interface, it's again converted back to string from boolean as in https://github.com/City-of-Helsinki/palvelutarjotin/pull/509/?

nikomakela (commented):
The new `isPartOfCulturalRoute` field in `EnrolmentForm` is now a bit hard to understand, which makes it also hard to maintain. Because of good unit tests, I think it's. still approvable.

src/domain/enrolment/enrolmentForm/__tests__/__snapshots__/EnrolmentForm.test.tsx.snap
karisal-anders: In the enrolment form the values are `""` (empty string) or either of the IsPartOfCulturalRoute enum's values i.e. IsPartOfCulturalRoute.NO_OR_UNKNOWN (i.e. `"NO_OR_UNKNOWN"` as a string value) or IsPartOfCulturalRoute.YES (i.e. `"YES"` as a string value).

The tests for `validIsPartOfCulturalRouteToBoolean` test that the enum values above pass and are converted to boolean values, and others like e.g. any of  [`'', ' ', 'yes', 'NO', 'unanswered', null, undefined, true, 0, 1]` raise an exception.

In the backend the value has slightly different meaning than in the frontend here, as in the backend there's also the possibility of "Unanswered" for old enrolments and queue enrolments made before this functionality has been deployed to production. On user input "Unanswered" is not allowed, but in the backend when converting `EnrolmentBase.is_part_of_cultural_route` boolean values the False value has "Unanswered", "No" and "I don't know" all in it, and conveying the reality of the information is IMHO good to have in reports so people interpreting the reports don't make false assumptions based on incorrectly presented data (If the "Unanswered" and "No / I don't know" cases saved as False would need to be separated it could still be done and that information conveyed by splitting the False values according to EnrolmentBase.enrolment_time to before deployment of this functionality and after).

src/domain/enrolment/enrolmentForm/EnrolmentForm.tsx
karisal-anders: I talked with the product owner about their needs and they wanted this so that people always choose a value, that they need to make a choice. So, making the user always choose a value for this is intentional and as per product owner's wishes.

